PR #15, opened 2018-12-12
Update readme to prevent link replacement

r1m:
Command lines links in the readme are replaced by <a> links.
This leads to commands rendered as `pip install git+reactiveops/reckoner` which obviously does not work.


ejether:
Thanks for the fix @r1m !

Issue #2, opened 2018-11-28
Add capability for additional args


ejether:
Carried over from: https://github.com/reactiveops/autohelm/issues/55

All in all, I think we could add a new multiple command line arg --helm-args, add a helm_args section to the course.yml spec and also per chart. As part of that, also deprecate the other command line args that are essentially just --helm-args but leaving the ones that require special logic. This would provide more than intended functionality, but also reduce the amount of changes required when new Helm args are released.

- [x]  add --helm-args to command line : https://github.com/reactiveops/autohelm/issues/54 
- [x] add helm_args section to course.yml
- [x] add deprecation warnings to  --debug flag
- [ ] remove --debug flag

Keep the existing reckoner --dry-run flag that will imply a helm dry-run and skip hooks.

sudermanjr:
By delegating `--dry-run` to helm-args, we would currently drop the functionality of being able to disable hooks running on `dry-run`.  Not sure we should do that.

ejether:
Hmm, not necessarily, because we could add a check for it in the `helm-arg` list but I'm not sure I like that better.

sudermanjr:
I think a helm dry-run and a reckoner dry-run become two separate things.  we can go ahead and remove debug (especially since dry-run implies a helm debug), but I think dry-run should remain.

Adding helm-args to course.yml is still a great idea though

ejether:
Would that make a reckoner dry run and helm dry run like:
`reckoner plot <course> --dry-run --helm-args=dry-run` our would reckoner `--dry-run` imply a helm `--dry-run`?

sudermanjr:
My thinking is the second.  reckoner dry-run implies a `helm --dry-run --debug` and also skips all hooks

PR #22, opened 2018-12-28
Addding the ability to specify global helm args in course.yml

sudermanjr:
Lots of small readability, formatting, white space, and typo fixes.
Specifying helm-args in cli overrides the course yaml file (with warnings)
Added helm_args to the test course.yml and added them to the test.
Warnings about deprecating --debug

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.9.0]
+
+### Added
+
+- the ability to specify helm_args in course.yml
+
+### Breaking changes
+
+- None, however the --debug flag will be deprecated in later versions. Warnings are added.
+
+### Fixed
+
+- numerous code formatting and spelling issues
+
 ## [0.8.4]
 
 ### Fixed 
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ helm plugin install https://github.com/reactiveops/reckoner
         `--helm-args --set=foo=toast`, or `--helm-args --recreate-pods`.
         Multiples are supported but only one parameter per `--helm-args` is
         supported. Note that specifying this flag will override `helm_args`
-        in the course.yml file.
+        in the course.yml file. WARNING: This cannot be used for arguments that specify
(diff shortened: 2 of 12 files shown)


sudermanjr:
Related to #2

tests/test_reckoner.py
```diff
@@ -271,7 +272,8 @@ def test_minimum_version(self):
         self.assertRaises(MinimumVersionException, self.c._compare_required_versions)
 
     def test_plot_course(self):
-        self.configure_subprocess_mock('', '', 0)  # Lots more work do do here with the installation of the list of charts
+        self.configure_subprocess_mock('', '',
+                                       0)  # TODO: Lots more work do do here with the installation of the list of charts
```
ejether: Is this formatting intentional?

sudermanjr: probably my IDE trying make PEP8 standards.  I can put it back if you want

ejether: Line length maybe? My pep8 formatter doesn't automatically fix that error. I don't really care, was just curious if it was an

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@ All notable changes to this project will be documented in this file.
...
 
+## [0.9.0]
+
+### Added
+
+- the ability to specify helm_args in course.yml
+
+### Breaking changes
+
+- None, however the --debug flag will be deprecated in later versions. Warnings are added.
```
endzyme: This is awesome - perfect way to address this upcoming breaking change!

README.md
```diff
@@ -36,6 +36,7 @@ helm plugin install https://github.com/reactiveops/reckoner
       * `--helm-args <helm-arg>`: Pass arbitrary flags and parameters onto helm
         such as `--helm-args --set=foo=toast`, or `--helm-args --recreate-pods`.
         Multiples are supported but only one parameter per `--helm-args` is supported.
+        Note that specifying this flag will override `helm_args` in the course.yml file.
```
endzyme: Not for this PR but something to actively think about as we move this forward:

We may want to define a common precedence for arguments defined in each "level" and the order of override. This will help the code use common implementations for which argument to keep or override

Example - 1 = lowest precedence and all higher numbers override
1. course.yml
2. environment variables
3. cli arguments

sudermanjr:
Turns out, this will not work with args that configure how helm connects to tiller.  This is due to the server version check at the beginning of the call.

TODO: We will document around this for the time being.

PR #26, opened 2019-01-03
Refactor the Helm Client

endzyme:
Refactor and Testing Augmentations

Refactored the helm client commands to allow for better isolation
between classes and easier unit testing. Also refactored some of the
dependent classes to consume a helm client vs instantiate one. This will
hopefully increase the ability to test each component for subtle bugs.

See Changelog for more information.


endzyme:
Also I tracked down that rollback doesn't actually run (see https://github.com/reactiveops/autohelm/blob/v0.7.2/autohelm/course.py#L116 from original refactor)

CHANGELOG.md
```diff
@@ -4,12 +4,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [0.9.1]
```
ejether: I think a rebase will fix this change

reckoner/chart.py
```diff
@@ -37,10 +37,10 @@ class Chart(object):
 
     Arguments:
     - chart (dict):
+    - helm: Instance of HelmClient()
```
ejether: OOC, why pass an instance instead of creating one or using a singleton?

tests/test_reckoner.py
```diff
@@ -342,61 +343,66 @@ def test_chart_install(self):
...
-                        '--recreate-pods',
-                        '--version=0.1.0',
-                        '--set-string=string=string',
-                        '--set-string=integer=10',
-                        '--set-string=boolean=True'
-                    ]
-                )
+            # TODO - we really need to refactor this to be better about testing in the same layer
+            #        this is traversing many layers in the code that could be better encapsulated
+            # COMMENTED OUT NICK HUANCA 2018-
```
ejether: Are you planning on refactoring this as part of this PR?

endzyme: I don't think so - it would be too large of a change to refactor if i'm not really changing too much here I wanted to keep all these tests as similar as possible.

CHANGELOG.md
endzyme: ah thanks - good catch

tests/test_reckoner.py
ejether: I agree that it would make it too large.
Is it worth leaving these tests in until you are ready to refactor them though? I'm not arguing that they don't need fixing, but they do currently serve to test the process in some way do they not?

reckoner/chart.py
endzyme: By passing one in you can use it as a ducktype interface so that testing the class in isolation can be done easier with mocks, without lots of monkeypatching. I can show you an example in slack if you want to talk through it more. This may not be pythonic but I'd love to see any references to do this more cleanly without having to instantiate mutliple classes to test one method in one class.

tests/test_reckoner.py
endzyme: I think they provide some value for sure. Even though it can be difficult to augment them right now, they provide a useful guide for what's valuable to test for functionality.

reckoner/chart.py
ejether: That makes sense. I honestly don't know what the most pythonic way to address this is.

endzyme:
RE: Smaller PRs vs all work being dependent -- Is there a particular piece you though as being out of scope? 

FYI some context on the change:
I was having trouble writing tests to isolate behavior in a class or subset of methods. I started down this road by removing the smallest common denominator I could re-structure without having to change all behavior. I thought the helm client was the easiest class to work with.

I think since I'm refactoring a core component which has lots of integrations across classes, it requires lots of file changes. Not sure if there is a mid-way step for completely replacing classes. Any help would be appreciated.

ejether:
I think the extra layers of isolation are good. I can see how much needs to change when you start moving things around in the core. 
I don't think this PR needs to be smaller but I wouldn't add any more concepts to it, if that makes sense?

endzyme:
going to be rebasing and pushing up FYI (in case anyone has this branch locally)

endzyme:
@ejether as it stands today, the rollback functionality in `reckoner` doesn't actually fire -- is it ok to hold off on the rollback implementation for another PR?

ejether:
That is unfortunate. Yes, I think that is probably another PR waiting to happen.

reckoner/helm/client.py
```diff
@@ -0,0 +1,159 @@
...
+        self._default_helm_arguments = value
+
+    def execute(self, command, arguments=[], filter_non_global_flags=False):
+        """
+        Run the command with the help of the provider.
+
+        return HelmCmdResponse
+        """
+
+        default_args = list(self.default_helm_arguments)
```
sudermanjr: I'm getting tripped up here when I don't pass any helm_args:
```
[ do-sfo2-sudermanjr   ] [pw] master  13s
 1 ✘± reckoner --log-level=DEBUG  plot course.yaml --dry-run
2019-01-03 15:05:57 root[83451]    DEBUG helm help
2019-01-03 15:05:57 root[83451]    DEBUG helm version --short --server
2019-01-03 15:05:58 root[83451]    DEBUG Installing repository: {'url': 'https://kubernetes-charts-incubator.storage.googleapis.com', 'name': 'incubator'}
Traceback (most recent call last):
  File "/Users/asuderma/.pentagon/pw/bin/reckoner", line 11, in <module>
    load_entry_point('reckoner', 'console_scripts', 'reckoner')()
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/asuderma/.pentagon/pw/lib/python2.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/cli.py", line 53, in plot
    h = Reckoner(file=file, dryrun=dry_run, debug=debug, helm_args=helm_args, local_development=local_development)
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/reckoner.py", line 72, in __init__
    self.course = Course(file)
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/course.py", line 72, in __init__
    repo.install()
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/repository.py", line 75, in install
    if self.name not in self._helm_client.repositories:
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/helm/client.py", line 64, in repositories
    raw_repositories = self.execute('repo', ['list'], filter_non_global_flags=True).stdout
  File "/Users/asuderma/repos/reactiveops/reckoner/reckoner/helm/client.py", line 35, in execute
    default_args = list(self.default_helm_arguments)
TypeError: 'NoneType' object is not iterable
```

endzyme: looking into it

endzyme: Found the problem! Thanks for tripping over it - I was exclusively testing with `test_course.yml` and not a very basic implementation. One thing for the list is end to end testing with simple and complex course.yml files.

sudermanjr: Awesome.  Works now.

Issue #18, opened 2018-12-20
FEATURE - Display a message when no chart matches --heading


sudermanjr:
When specifying a `--heading`, it would be nice if Reckoner told you that there was no release matching that name.

endzyme:
I'll add this to #56

endzyme:
> I'll add this to #56

Nope actually skipping this addition for now

endzyme:
#63 resolves this

PR #56, opened 2019-02-07
WIP Reckoner 0.13.0

endzyme:
### Fixes
* `version` command works again
* Non-working `generate` subcommand is removed
* Added functionality to fail and display help when you just run `reckoner` with no subcommand
* Added tests to some of the contracts in the cli.py

### Changes
* Initial support for migration from `values: {}` to `sets: {}`. See #7 for details.

### DIFF
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -221,10 +221,10 @@ class Chart(object):
             [logging.warning(msg) for msg in self._deprecation_messages]
 
     def build_helm_arguments_for_chart(self):
-        '''
+        """
         This method builds all the arguments we'll pass along to the helm
         client once we need to run the install
-        '''
+        """
 
         # Set Default args (release name and chart path)
         self.args = [
@@ -268,19 +268,19 @@ class Chart(object):
         self.build_set_string_arguments()
 
     def _merge_set_and_values(self):
-        '''
+        """
         This is a temporary method that will be gone once the values: vs set:
         debacle has been resolved. (See https://github.com/reactiveops/reckoner/issues/7)
 
         NOTE ONLY RUN THIS ONCE BECAUSE IT'S NOT IDEMPOTENT
-        '''
+        """
         if self._hack_set_values_already_merged:
             raise StandardError('This method cannot be called twice. '
                                 'If you are seeing this please open an '
                                 'issue in github.')
 
         def merge_dicts(values, sets):
-            '''This does a dict merge and prefers "sets" values'''
+            """This does a dict merge and prefers "sets" values"""
             new_dict = values.copy()
             new_dict.update(sets)
             return new_dict
@@ -292,12 +292,12 @@ class Chart(object):
         self._hack_set_values_already_merged = True
 
     def build_temp_values_files(self):
-        '''
+        """
         This method builds temporary files based on the values: settings
         provided in the course.yml. This effectively keeps type persistence
         between the course.yml and what is passed to helm. If you use set:
         value arguments then you can lose types like int, float and true/false
-        '''
+        """
         if self.values:
             self._deprecation_messages = [
                 "DEPRECATION NOTICE: The behavior of `values: {}` will change in "
@@ -317,33 +317,33 @@ class Chart(object):
             )
 
     def build_files_list(self):
-        '''
+        """
         This method builds the files list for all
         files specified in the course.yml
-        '''
+        """
         for values_file in self.files:
             self.args.append("-f={}".format(values_file))
 
     def build_set_string_arguments(self):
-        '''
+        """
         Builder for "set-string" arguments in helm command line
         This method specifically modifies the Chart object to
         prepare the command line arguments.
 
         Note running this multiple times will provide duplicate arguments
-        '''
+        """
         for key, value in self.values_strings.iteritems():
             for k, v in self._format_set(key, value):
                 self.args.append("--set-string={}={}".format(k, v))
 
     def build_set_arguments(self):
-        '''
+        """
         Builder for "set" arguments in helm command line
         This method specifically modifies the Chart object to
         prepare the command line arguments.
 
         Note running this multiple times will provide duplicate arguments
-        '''
+        """
         for key, value in self.set_values.iteritems():
             for k, v in self._format_set(key, value):
                 self.args.append("--set={}={}".format(k, v))
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -67,6 +67,8 @@ class Chart(object):
         if value_strings != {}:
             del(self._chart['values-strings'])
 
+        self._deprecation_messages = []
+
     @property
     def helm_args(self):
         """ Returns list of extra options/args for the helm command """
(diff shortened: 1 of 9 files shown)


reckoner/chart.py
```diff
@@ -164,12 +160,13 @@ def update_dependencies(self):
         """ Update the course chart dependencies """
         if self.config.local_development or self.config.dryrun:
             return True
-        logging.debug("Updating chart dependencies: {}".format(self.chart_path))
-        if os.path.exists(self.chart_path):
```
endzyme: I have refactored self.chart_path because it wasn't used anywhere external  and just leveraged repository chart path directly.

reckoner/chart.py
```diff
@@ -132,13 +128,13 @@ def post_install_hook(self):
 
     def run_hook(self, hook_type):
         """ Hook Type. Runs the commands defined by the hook """
-        coms = self.__get_hook(hook_type)
-        if coms is None:
-            return coms
-        if type(coms) == str:
-            coms = [coms]
+        commands = self.__get_hook(hook_type)
```
endzyme: This was a linting change. Comms to commands

reckoner/chart.py
```diff
@@ -178,9 +175,6 @@ def install(self, namespace=None, context=None):
 
         Arguments:
         - namespace (string). Passed in but will be overridden by Chart().namespace if set
-
```
endzyme: I've updated documentation here because this method didn't actually return a bool value

reckoner/chart.py
```diff
@@ -191,63 +185,162 @@ def install(self, namespace=None, context=None):
...
-                self.args.append("--set={}={}".format(k, v))
 
+        # HACK: This is in place until we can fully deprecate the usage of set
+        #       in place of values. Currently values: gets translated into
+        #       --set commands to the helm command line run. This should be
+        #       changed so that set: goes to command line args and values: goes
+        #       to temporary values files, to keep strong types in yaml.
+        #       This will go away once we fully change the functionality of
+        #       values: vs sets: settings in course.yml
+        self._merge_set_and_values()
```
endzyme: This is the main function that enables the migration path between values and sets. the idea is that we set all values as sets and then at some point in the future we transition it.

reckoner/chart.py
```diff
@@ -191,63 +185,162 @@ def install(self, namespace=None, context=None):
...
+        self.build_set_string_arguments()
+
+    def _merge_set_and_values(self):
+        '''
+        This is a temporary method that will be gone once the values: vs set:
+        debacle has been resolved. (See https://github.com/reactiveops/reckoner/issues/7)
+
+        NOTE ONLY RUN THIS ONCE BECAUSE IT'S NOT IDEMPOTENT
+        '''
+        if self.__already_merged:
```
endzyme: This is lazy naming and should be something like `__hack_merged_values_sets`

reckoner/chart.py
```diff
@@ -191,63 +185,162 @@ def install(self, namespace=None, context=None):
...
+        value arguments then you can lose types like int, float and true/false
+        '''
+        logging.warning(
+            "DEPRECATION NOTICE: The behavior of `values: {}` will change in "
+            "v0.14+. Currently values: are set as --set arguments for the "
+            "helm command. This behavior makes all values: you set in the "
+            "course.yml lose type fidelity, true becomes string(\"true\") "
+            "and all int values become a string of the integer. See "
+            "https://github.com/reactiveops/reckoner/issues/7 for details."
+        )
```
endzyme: This should have more instructional content. something that says change your values key two sets to keep the same behavior.

reckoner/tests/test_cli.py
```diff
@@ -0,0 +1,59 @@
...
+        result = runner.invoke(cli.cli, args=['--version'])
+
+        self.assertEqual(0, result.exit_code)
+        self.assertNotEqual('', result.output)
+
+    def test_supports_loglevel(self):
+        '''assure we have a global command for --log-level'''
+        runner = CliRunner()
+        result = runner.invoke(
+            cli.cli, args=['--log-level', 'DEBUG', 'version'])
```
endzyme: This test kinda sucks, there's probably a better way to aggregate available parameters.

reckoner/tests/test_cli.py
```diff
@@ -0,0 +1,59 @@
...
+
+        self.assertEqual(0, result.exit_code)
+        self.assertNotEqual('', result.output)
+
+    def test_exits_without_subcommad(self):
```
endzyme: Typo subcommad

tests/test_reckoner.py
```diff
@@ -19,8 +19,6 @@
 import coloredlogs
 import logging
 import os
-import git
-import subprocess
```
endzyme: I've been removing unused imports

endzyme:
Make sure to add #18

sudermanjr:
If we are going to change values to set-values, would it be best for consistency to also changes `values-strings` to `set-values-strings` ?

Also, are we planning to bump the major version when we actually proceed with the deprecation?

endzyme:
I would think we could either support both `values-strings` and `set-values-strings` until the refactor and give a warning notice about the deprecation. 

In terms of the major version bump, not really sure. Are we good to jump to 1.x? How stable are we feeling today D:

ejether:
💡 
What about a nested structure instead of long verbose keys? 
```
set: #structre that adds --set=
  values: ~
  strings: ~
values: ~ # structure that creates temp file
```

sudermanjr:
> In terms of the major version bump, not really sure. Are we good to jump to 1.x? How stable are we feeling today D:

I'm feeling pretty stable about this particular project.  Once we get these changes in it will be a very serious upgrade in functionality and stability.

ejether:
I agree. With this change it feels "complete" and 1.x ready

reckoner/chart.py
```diff
@@ -191,63 +185,168 @@ def install(self, namespace=None, context=None):
...
+        logging.debug('New value of sets: {}'.format(self.set_values))
+        self._hack_set_values_already_merged = True
+
+    def build_temp_values_files(self):
+        '''
+        This method builds temporary files based on the values: settings
+        provided in the course.yml. This effectively keeps type persistence
+        between the course.yml and what is passed to helm. If you use set:
+        value arguments then you can lose types like int, float and true/false
+        '''
```
sudermanjr: nit: several places use `'''` instead of `"""`.  Not sure if intentional.

reckoner/chart.py
```diff
@@ -191,63 +185,168 @@ def install(self, namespace=None, context=None):
...
+        if self.values:
+            self._deprecation_messages = [
+                "DEPRECATION NOTICE: The behavior of `values: {}` will change in "
```
sudermanjr: I'm not certain we need quite this much info in the message.  As a user, I find overly long deprecation messages annoying and tend to skip over them.  The link to the issue is really good and will give most of this information to the user.

reckoner/chart.py
endzyme: So the `"""` are from code i'd moved around - I had been using `'''` - I'd actually not notices that I had used the opposite that the other files had used. I can refactor but there [isn't really any difference](https://stackoverflow.com/questions/7783097/when-to-use-triple-single-quotes-instead-of-triple-double-quotes) in the implementation of what it does.

reckoner/chart.py
endzyme: Ok - I can summarize this a bit by using `"DEPRECATION NOTICE: Change `values: {}` to `set-values: {}` to keep consistent behavior beyond v1.x. Details : github.com/....."`

Sound ok?

reckoner/chart.py
sudermanjr: As long as we are consistent, I don't care, and I know it's not different.  Thus the `nit` :-D

endzyme:
Re: the nested set values and strings yaml blocks. I'm worried too much insetting will lead to confusion, and since we're not running this through a schema validator, it's easy to unset it inadvertently lose settings silently. 

I'm all for added structure depth but in this scenario it feels like a little too much. I was also hoping for a simple deprecation message for people to read and implement, and the added depth just feels a little more confusing than a key rename.

reckoner/chart.py
sudermanjr: That sounds awesome.

ejether:
> Re: the nested set values and strings yaml blocks. I'm worried too much insetting will lead to confusion, and since we're not running this through a schema validator, it's easy to unset it inadvertently lose settings silently.
> 
> I'm all for added structure depth but in this scenario it feels like a little too much. I was also hoping for a simple deprecation message for people to read and implement, and the added depth just feels a little more confusing than a key rename.

 👍 I was just thinking outloud.

endzyme:
Closing to start a new PR for v1.0.0-rc1

PR #58, opened 2019-02-07
Updating readme.  Fixes #57

### LINKED ISSUE #57: README Fixes
The README still suggests installing with pip directly from git, and it still has the helm plugin as an option.  We should remove these and stick to just `pip install reckoner`

sudermanjr:


sudermanjr:
> Also remove `plugin.yaml` if you're backing away from reckoner being a helm plugin

@ejether done.

PR #61, opened 2019-02-08
Release Candidate for 1.0.0 (RC1)

endzyme:
## Changes
* Adjusted naming of hack_merge variable
* Added information to the deprecation warning
* Adjusted the deprecation to be fired last after the final hook (so it is easily noticed)
* More tests to assure we avoid breaking changes to the CLI params and arguments

## Features
* A new `set-values: {}` key in the chart config to use for settings `--set` for helm

## Deprecation Notice
In the future, the `values: {}` configuration of a chart will change it's
behavior (see Issue #7). The current behavior translates any `values:{}` into
`--set` for the command line helm. This makes certain object types lose their
type fidelity. This means `true` becomes `string("true")` when pushed through
`--set values=true`. To maintain the same behavior for your settings please
change `values:{}` to `set-values:{}`. This behavior deprecaton will start to
be enforces in later versions of Reckoner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.0.0-rc1]
+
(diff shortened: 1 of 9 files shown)


sudermanjr:
Previous discussion located here: https://github.com/reactiveops/reckoner/pull/56

PR #63, opened 2019-02-09
Fixed error handling

endzyme:
Reckoner now raises an error to the user and is handled gracefully by click.

* Added error handling to click plot command to expect reckoner exceptions
* Fixed inheritance of reckoner exceptions
* Added tests around this new behavior
* Added testing to assure if at least a single chart in --only is correct, the run still succeeds but warns you at the end of missing charts
* Removed mention of "rollback" - because it hasn't worked since the reckoner rename

### DIFF
--- a/reckoner/cli.py
+++ b/reckoner/cli.py
@@ -15,14 +15,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import logging
 import coloredlogs
 import click
-import shutil
-from reckoner import Reckoner
-import pkg_resources
-
-from meta import __version__
+from .reckoner import Reckoner
+import exception
+from .meta import __version__
 
 
 @click.group(invoke_without_command=True)
@@ -53,8 +50,14 @@ def cli(ctx, log_level, *args, **kwargs):
                                                                        'development.')
 def plot(ctx, file=None, dry_run=False, debug=False, only=None, helm_args=None, local_development=False):
     """ Install charts with given arguments as listed in yaml file argument """
-    h = Reckoner(file=file, dryrun=dry_run, debug=debug, helm_args=helm_args, local_development=local_development)
-    h.install(only)
+    try:
+        h = Reckoner(file=file, dryrun=dry_run, debug=debug, helm_args=helm_args, local_development=local_development)
+        # Convert tuple to list
+        only = list(only)
+        h.install(only)
+    except exception.ReckonerException:
+        # This handles exceptions cleanly, no expected stack traces from reckoner code
+        ctx.exit(1)
 
 
 @cli.command()
--- a/reckoner/course.py
+++ b/reckoner/course.py
@@ -21,11 +21,11 @@ import sys
 
 import oyaml as yaml
 
-from config import Config
-from chart import Chart
-from repository import Repository
-from exception import MinimumVersionException, ReckonerCommandException
-from helm.client import HelmClient, HelmClientException
+from .config import Config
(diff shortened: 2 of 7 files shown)


reckoner/reckoner.py
```diff
@@ -85,7 +86,23 @@ def install(self, charts=[]):
 
         """
         selected_charts = charts or [chart._release_name for chart in self.course.charts]
-        return self.course.plot(selected_charts)
+        try:
+            self.course.plot(selected_charts)
+        except NoChartsToInstall as error:
```
endzyme: probably could use a test here to assure this function can handle this exception

tests/test_reckoner.py
```diff
@@ -95,6 +96,51 @@ def test_install_succeeds(self, *args):
...
+        course_yml = {
+            'namespace': 'fake',
+            'charts': {
+                'fake-chart': {
+                    'chart': 'none',
+                    'version': 'none',
+                    'repository': 'none',
+                }
+            }
+        }
```
endzyme: This is an example of how we could get rid of yaml files in the tests

reckoner/course.py
```diff
@@ -115,6 +114,19 @@ def plot(self, charts_to_install):
...
+                    'Skipping {} in course.yml, not found '
+                    'in your requested charts list'.format(chart.release_name)
+                )
+
+        if len(self._charts_to_install) == 0:
+            raise NoChartsToInstall(
+                'No charts found from requested list ({}). They do not exist '
+                'in the course.yml. Verify you are using the release-name '
+                'and not the chart name.'.format(', '.join(charts_to_install))
+            )
```
endzyme: This could be a little confusing and might be better suited as a set of method. Basic concept is that for each chart we've asked to install, we'll add it to `self._charts_to_install` and remove it from `charts_to_install` (yes confusing I know, didn't want to refactor all the existing names of variables, really this should be called `requested_release_names`). 

Once we find that there is nothing in `self._charts_to_install` then we explode an error to the user.

endzyme:
How to test this

* Run a normal install no --only or --heading
* Run an install with two non-existent --heading definitions
* Run an install with a non-existent --heading definition **and** a valid --heading definition
* Run an install with one non-existent --heading definition

reckoner/cli.py
```diff
@@ -53,8 +50,14 @@ def cli(ctx, log_level, *args, **kwargs):
                                                                        'development.')
 def plot(ctx, file=None, dry_run=False, debug=False, only=None, helm_args=None, local_development=False):
     """ Install charts with given arguments as listed in yaml file argument """
-    h = Reckoner(file=file, dryrun=dry_run, debug=debug, helm_args=helm_args, local_development=local_development)
-    h.install(only)
+    try:
+        h = Reckoner(file=file, dryrun=dry_run, debug=debug, helm_args=helm_args, local_development=local_development)
+        # Convert tuple to list
+        only = list(only)
```
ejether: Is an explicit conversion here the right thing to do or should the class handle poor input gracefuly by raising and error?

endzyme: Could go either way - this is why I prefer stronger typed languages :) -- python3 will get us type hinting at least - so something to look forward to.

endzyme: Do you have a preference? I can change if you do - It really doesn't matter to me either way

endzyme:
I am going to merge this and we can reconvene on some of the details before we go to master.

Issue #59, opened 2019-02-08
Errors in Post Install Hook Are Not Logged


robscott:
It seems like whatever output comes from a post install hook doesn't actually make it to stdout. That means if you happening to be referencing an invalid path, there will be nothing in the Reckoner output to indicate there's a problem with your config.

endzyme:
Quick clarification - would you expect it to always return stderr or only when exitcode != 0

endzyme:
Other questions on expectations

Today all hooks just blindly accept that they can fail but continue with the chart installation. 

1. should pre-install hooks stop a chart from running
2. should post-install hooks stop the next chart from running
3. if there are three hooks in pre-install, and hook 2 fails, should hook 3 run?

TL;DR; what should reasonable error handling look like for hooks?

PR #64, opened 2019-02-12
Fixing issue 59

endzyme:
#59 is addressed in this PR

Behavior today was only showing output of hooks, regardless of failure, in the debug output. We fixed up the handling of errors to match what the call() function returns. This now checks for exit success and will log stderr and stdout if there is anything to bubble up.


endzyme:
This PR Requires #58  first!

Issue #60, opened 2019-02-08
Post Install Hooks Should Run Relative to Course.yml Path


robscott:
This would help make reckoner behavior more consistent for teams that may not consistently be running a `reckoner plot` from the same path.

ejether:
💡 Made me think about the concept of a `workingpath` for the hooks with the default being `os.path.dirname(course.yml)`

PR #66, opened 2019-02-14
Run hooks from course.yml directory

endzyme:
This changes the behavior to run the hooks from the course.yml directory by default. You can still define absolute paths, but all relative paths should start from where course.yml exists. This is how I understand most implementation are today (either absolute or relative to course.yml).


endzyme:
requires #64

endzyme:
fixes #60

PR #69, opened 2019-02-15
Reckoner v1.0.1

endzyme:
Fixes
* #60 - Better post hook experience with relative pathing to course.yml
* #59 - Better visibility into post hooks
* #18 - Resolved missing heading definitions with better error handling

PR #70, opened 2019-02-16
Fix hook execution path

ejether:
The dirname() was returning "" empty white space which was causing subprocess to try and cd to "" dir which cannot exist. This may not have been caught if ./ was used in testing

### DIFF
--- a/reckoner/config.py
+++ b/reckoner/config.py
@@ -70,10 +70,7 @@ class Config(object):
 
     @property
     def course_base_directory(self):
-        if self.course_path:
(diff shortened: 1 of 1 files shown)


endzyme:
Interesting, can you post how to test this?

sudermanjr:
See my snippet in #69.  This addresses that issue.

endzyme:
Ah I see. Looking at docs for dirname - if you run from the same dir as course.yaml it will set the cwd to `""`. Indeed I only manually tested the new behavior. Nailed it.

Issue #76, opened 2019-03-13
A Failing Pre-Install hooks should block chart install 


endzyme:
Currently the behavior of reckoner is to continue on if any of the pre or post installation hooks fail. This can be a confusing user experience, especially if you've fired 20 chart installs and the first one had a failure. This signal can be lost in the noise of output from Reckoner. 

Feature Request
- Could Reckoner, by default, stop after the first failed pre-install hook and notify the customer.
- Could we add a quick summary of failed chart installations at the end of the run (failed chart install N shouldn't block chart N+1)
- Could we add an override to ignore all failed pre-install-hooks
- Failed post-install hooks should also be added to the summary list of failed chart installs as "Failed  Post-Install"
- All hooks should stop running after the first one fails (example below)

```yaml
charts:
  my-chart:
    set-values:
      enabled: true
    hooks:
      pre_install:
        - echo true
        - curl non*possible().domain
        - echo true   #<--- this should never fire if the curl above fails & chart will also not be installed
      post_install:
        - echo true
        - cat fileno-existy
        - echo this-never-runs-in-post
```

ejether:
Ah, I think the change @sudermanjr created was that if the hook fails, the chart won't be installed.
This is so that if any hook command fails, the rest of the hooks should not be run.

endzyme:
Do you have that PR handy? Was this in the previous version before the rename? I am curious if I clobbered that behavior accidentally with my changes recently. I think this behavior should have a test for sure and I'll see if I can make some time to test it soon.

ejether:
I'll see if I can dig it up. I'm 99% sure we had it exit on failure of a hook.

endzyme:
Yup ok - so I traced back behavior on and old ref. 

1. Calls install on a chart definition https://github.com/reactiveops/reckoner/blob/307e66b551386e82406d84f90449a3908caed2ef/reckoner/course.py#L119 
2. Install runs pre_install hook https://github.com/reactiveops/reckoner/blob/307e66b551386e82406d84f90449a3908caed2ef/reckoner/chart.py#L194
3. run_hook runs the command and raises an error if exitcode > 0 (https://github.com/reactiveops/reckoner/blob/307e66b551386e82406d84f90449a3908caed2ef/reckoner/chart.py#L150-L153 & https://github.com/reactiveops/reckoner/blob/307e66b551386e82406d84f90449a3908caed2ef/reckoner/__init__.py#L84-L85)
[...]

PR #77, opened 2019-03-15
Fixing broken hook runs

endzyme:
Overview of the breaking change I made here.
1. I changed the contract of the call() function
   so it always returns a response even if the
   exitcode was not 0. (this was a change from before)
   The intent was so that we can run commands and decide
   what to do upstream depending on exitcode or other
   information.
2. The tests I wrote actually should have been bubbling up
   errors from the run_hook() method, since we are not
   throwing the error at call().

I have added tests to make sure we throw errors with
non-zero exit codes.

### DIFF
--- a/reckoner/tests/test_course.py
+++ b/reckoner/tests/test_course.py
@@ -103,9 +103,10 @@ class TestIntegrationWithChart(unittest.TestCase):
             }
         }
 
-        chartCallMock.side_effects = [Response(exitcode=1, command_string='mocked', stderr=' ', stdout=' ')]
+        chartCallMock.return_value = Response(exitcode=1, command_string='mocked', stderr=' ', stdout=' ')
 
         course = Course(None)
         course.plot(['first-chart'])
 
+        self.assertEqual(chartCallMock.call_count, 1)
         self.assertEqual(len(course.failed_charts), 1, "We should have only one failed chart install due to hook failure.")
--- a/reckoner/tests/test_course.py
+++ b/reckoner/tests/test_course.py
@@ -14,8 +14,8 @@
 
 import mock
 import unittest
-from reckoner.exception import MinimumVersionException
 from reckoner.course import Course
+from reckoner.command_line_caller import Response
 
 
 @mock.patch('reckoner.repository.Repository', autospec=True)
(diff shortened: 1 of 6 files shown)


endzyme:
Fixes #76

reckoner/tests/test_course.py
```diff
@@ -73,3 +73,39 @@ def test_init_error_fails_min_version_helm(self, configMock, helmClientMock, yam
...
+                    'hooks': {
+                        'pre_install': [
+                            'run a failed command here',
+                        ]
+                    }
+                }
+            }
+        }
+
+        chartCallMock.side_effects = [Response(exitcode=1, command_string='mocked', stderr=' ', stdout=' ')]
```
sudermanjr: Trying to check this testing.  If I change the exitcode to 0 for this, then the test should fail, correct?

endzyme: hmmm well depends on if everything else in the `.plot()` succeeds. It's not a very effective integration test since there are so many moving pieces between `course.plot()` and `chart.run_hooks().call()`. Have you tried it and the negative testing didn't do what you expected?

endzyme: > I tested the intended functionality, looks good. My only issue (that doesn't necessarily need to be addressed in this PR), is the feedback to the user. It does a great job of outputting the error from the hook, but then there is no indication that the install was skipped because of that. It just says "it failed", which leads the user to believe that the helm install failed. It would be nice to have a "Reason for chart failure" as part of the message there instead of just `Helm upgrade failed on rbac-manager`

This is something I can definitely augment if you'd like. I think it could be something tucked into the class instance of the chart that failed. Maybe another property for "reason failed" so that we can aggregate that at the end of the run.

This addition would be a net new from fixing the bug I originally introduced (correct me if I'm wrong). Since it would be net new we could open an issue for it and pick it up as a future augmentation. I think it would be valuable to be able to see what caused chart failures in general, not only if it's just a pre-install hook.

endzyme:
> I tested the intended functionality, looks good. My only issue (that doesn't necessarily need to be addressed in this PR), is the feedback to the user. It does a great job of outputting the error from the hook, but then there is no indication that the install was skipped because of that. It just says "it failed", which leads the user to believe that the helm install failed. It would be nice to have a "Reason for chart failure" as part of the message there instead of just `Helm upgrade failed on rbac-manager`

This is something I can definitely augment if you'd like. I think it could be something tucked into the class instance of the chart that failed. Maybe another property for "reason failed" so that we can aggregate that at the end of the run.

This addition would be a net new from fixing the bug I originally introduced (correct me if I'm wrong). Since it would be net new we could open an issue for it and pick it up as a future augmentation. I think it would be valuable to be able to see what caused chart failures in general, not only if it's just a pre-install hook.

sudermanjr:
> This addition would be a net new from fixing the bug I originally introduced (correct me if I'm wrong). Since it would be net new we could open an issue for it and pick it up as a future augmentation. I think it would be valuable to be able to see what caused chart failures in general, not only if it's just a pre-install hook.

Definitely agree this should be a new issue and a new PR.  I was thinking along the same lines as you (adding a new property to that class).  I'll go ahead and open the issue.

reckoner/tests/test_course.py
sudermanjr: I tried changing the exit code, and  the test passed.  However, if I leave the exit code at 0 and change the expected number of failed charts to 0, the test fails as expected.  

I agree that's a tough testing scenario, but after further playing around I think this is fine for what you are trying to accomplish.

endzyme: Actually looks like this test was broken :D  -- thanks for finding this. It's now fixed. The side_effect wasn't passing the right object. Not sure why still but really this whole part of the code needs some loves to build in better isolation.

PR #79, opened 2019-03-18
Pytest change

endzyme:
Pipeline is currently broken because nose2 sometimes doesn't load the mock (WAT?)

I am going to be retagging this merge as v1.0.2 due to the problems with the pipeline currently. 

How to reproduce the error:
```
# from root directory
nose2   # This fails same as CI does
nose2 -s reckoner/tests # This passes and it runs the test that fails in the previous command
pytest # always works 
```
When debugging i could see that the mock wasn't being instantiated with nose2 when running from root. When you add a start directory to nose2, it *does* use the mock, which is super weird and inconsistent. 

I'm opting to move this all to pytest instead of figuring out why nose2 isn't working.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -28,7 +28,7 @@ jobs:
             sudo pip install .
             reckoner --version
             sudo pip install -r tests/requirements.txt
-            nose2
+            pytest
       - setup_remote_docker
       - run:
           name: Build Docker Image
@@ -84,5 +84,3 @@ workflows:
               only: /.*/
             branches:
               ignore: /.*/
-
-
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,1 +1,2 @@
 pytest
+mock
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,2 +1,1 @@
-nose2==0.8.0
 pytest
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,1 +1,2 @@
 nose2==0.8.0
+pytest


ejether:
Are the tests still duplicated between the root `/tests` and `/reckoner/tests` and other locations? 
What was the reasoning for interspersing the tests throughout the module in the first place, I'm not clear on why that pattern.

endzyme:
> Are the tests still duplicated between the root `/tests` and `/reckoner/tests` and other locations?
> What was the reasoning for interspersing the tests throughout the module in the first place, I'm not clear on why that pattern.

woops - damn sorry i missed your comment here.

/tests/ and /reckoner/tests/ definitely need some more definition. I started putting things into `/reckoner/tests` due to the staged assets in the other file which probably could be broken down into different files. The main reason was that using the large `/tests/` single file was confusing. I could definitely use some attention to bridge the gap

ejether:
Ah, I see. Breaking up the tests file makes a lot of sense. Typically the patter I've seen is for that to be broken up in a `/tests/` directory so like:
```
.tests
├── __init__.py
├── charts.py
├── command_line_caller.py
└── reckoner.py
```
or 
```
tests
├── __init__.py
├── charts
├── cli
├── command_line_caller
└── reckoner
```

Is that a pattern you avoided intentionally?

endzyme:
Good question. My goals, in previous PRs, was to add more isolated testing and keep things covered. I hadn't been to refactor that test area. I'm not a python pro, so my googling led me to believe that tests specific to modules go into that package folder. I used that pattern for new tests. Even tests that span two classes (integration), since those files are in the same module.

In terms of breaking up the tests, I think that would still be prudent. I think before any of that work begins it would be good to layout goals and a structure to adhere to. I still think this package could use more end to end testing, and `/tests/` might be a good place for that.

ejether:
That makes sense. This gist sums up my understanding of the typical patterns for project organization WRT unit test files. https://gist.github.com/tasdikrahman/2bdb3fb31136a3768fac
Regardless, coverage is more important than a strict organization. I did find that without the top level `/tests` here, nose ran just fine so the root cause was probably a conflict at the various levels of organization though.

I don't have a strong sense of where e2e tests belong.

Thanks for satisfying my curiosity re: the test organization!

PR #83, opened 2019-03-27
Upgrade oyaml to >=0.8 and PyYAML to >=5.0

coreypobrien:


setup.py
```diff
@@ -40,9 +40,9 @@
       packages=find_packages(exclude=('tests','*.tests')),
       install_requires=[
         "click==6.7",
-        "PyYAML==4.2b1",
+        "PyYAML>=5.0",
```
endzyme: We can probably remove this because pip will pull in the right requirement from `oyaml`.

reckoner/course.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, course_file):
         Parse course.yml contents into instances.
         """
         self.config = Config()
-        self._dict = yaml.load(course_file)
+        self._dict = yaml.load(course_file, Loader=yaml.loader.FullLoader)
```
endzyme: I'm getting a deprecation notice in `tests/test_reckoner.py` due to not having the `Loader=` present on `yaml.load(...)`. Can we make sure the tests are updated too?

endzyme:
I think this change should also bump the internal version of the whole app itself. We can do that from master if you want when we want to bump, but could you put some info in the changelog?

endzyme:
@coreypobrien any ideas on if you'll be able to get to this in the near future?

PR #84, opened 2019-04-05
Python 3 Support

endzyme:
Hats off to @bambash for kicking this off!

This migration is intended to enable future deprecation of python 2.7 but also supporting python 3.6 & 3.7. 

This PR:
* Enables pytest being run on python 2.7, 3.6 and 3.7
* Changes in python code to support both 2 and 3

This is just adding visibility to non-python3 changes going forward in the library. It will also fail if something is added that breaks python2, but the idea is that we should move away from that asap.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,9 +15,49 @@
 version: 2
 
 jobs:
-  build:
+  build-2.7:
     docker:
-      - image: circleci/python:3
+      - image: circleci/python:2.7
+    environment: ~
+    working_directory: ~/reckoner
+    steps:
+      - checkout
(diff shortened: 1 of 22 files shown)


sudermanjr:
This looks super awesome.  I'm firing up a functional testing env for this right now.

endzyme:
Ok the last change fixes some issues with string comparisons in reckoner with python3. Stdout and Stderr from Popen().communicate() return as bytes in python 3 so comparing them to other strings loaded from yaml would have different behavior between versions.

We're having all calls return all stdout and stderr as string data now which should keep the behavior consistent.

PR #85, opened 2019-04-08
Versioning Changes

endzyme:
Reckoner versions now can be based on git tag. This also shows users who install via source if their repo has any different changes. This has some regex that makes the version more compatible with semver module for comparisons.

Main Benefits
- When developing / installing locally you can see if you've installed from a dirty branch or are ahead of a released version
- You now do releases by just tagging the git repo and pushing the tag, no more managing meta.py for `__version__`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,20 +61,51 @@ jobs:
     environment: ~
     working_directory: ~/reckoner
     steps:
+      - run:
+          name: Setup PATH to support pip user installs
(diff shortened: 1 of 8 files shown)


sudermanjr:
Did you pull this off the python3 branch?  It looks like it also includes all of that PR as well.

endzyme:
Yup. That's expected due to the large changes on circleci. I should have noted that more in the PR description

PR #87, opened 2019-04-10
Dynamic versioning

endzyme:
* Migrated the ci pipeline to use pip install --user
* Removed unneeded settings from circleci
* Removed tests/requirements.txt and added the items to development-requirements.txt
* Refactored the versioning to be driven by nearest git tag

Git tags now drive the pip version, which will:
* Help understand if you've installed from the source and have a dirty or modified version (it's listed in pip now with the commit sha if it's N commits away from nearest version)
* Help ease versioning burden by just tagging the release and not having to update meta.py or setup.py

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,20 +61,51 @@ jobs:
     environment: ~
     working_directory: ~/reckoner
     steps:
+      - run:
+          name: Setup PATH to support pip user installs
+          command: echo 'export PATH=$PATH:/home/circleci/.local/bin' >> $BASH_ENV
       - checkout
       - run:
           name: Unit Tests
           command: |
-            sudo pip install .
+            pip install --user -r development-requirements.txt
+            pip install --user -e .
             reckoner --version
-            sudo pip install -r tests/requirements.txt
             pytest
-      - setup_remote_docker
+  build-3.6:
+    docker:
+      - image: circleci/python:3.6
+    environment: ~
+    working_directory: ~/reckoner
+    steps:
       - run:
-          name: Build Docker Image
+          name: Setup PATH to support pip user installs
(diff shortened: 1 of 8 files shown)


ejether:
The PR is fine but I'm not clear what value this work provides. 
Can you provides some background information on the 'why' here?

endzyme:
Sure! For which part specifically? Above I stated why I'm interested in dynamic versioning and what value it provides. As for the rest I'll outline below with more context.

* Migrated the ci pipeline to use pip install --user 
>I'm just used to installing things as user instead of leveraging sudo in the container. 

* Removed unneeded settings from circleci
>We were setting environment and also building docker images what weren't used at all so I removed them

* Removed tests/requirements.txt and added the items to development-requirements.txt
>This was a decision to have less requirement texts that are spread around. It was a problem I introduced initially.

endzyme:
@ejether I wanted to honor your questions before merging this just to make sure the intent is captured.

.circleci/config.yml
```diff
@@ -92,13 +79,6 @@ jobs:
             echo -e "[pypi]" >> ~/.pypirc
             echo -e "username = $PYPI_USERNAME" >> ~/.pypirc
             echo -e "password = $PYPI_PASSWORD" >> ~/.pypirc
-      - run:
-          name: create release
```
sudermanjr: Are we no longer automatically creating a github release?  Is this something we will have to do manually from here out?

sudermanjr: I can't remember if that was even working, mostly just curious.

endzyme: The release part hasn't worked in a while (for me at least) - I have been manually creating the release in Github.

endzyme: https://github.com/reactiveops/reckoner/issues/30

sudermanjr: Thanks.  Couldn't remember

Issue #86, opened 2019-04-09
Feature request "--strict" mode


willnewby:
I notice that Reckoner when it fails to install/upgrade a chart, just continues on. I'd really like a strict mode that bails out the whole run and returns an exit code 1 (or something non-zero) if any chart fails. My use case here is a CI/CD flow, since my jobs all pass even if some of the charts fail (which means errors aren't surfaced).

Thanks!

endzyme:
@willnewby I'm picking this up now and I wanted to know if we can do two things.

1. The new behavior is to always exit code != 0 when a chart or hook fails to run, but would still execute all charts in the course definition
1. A new flag of `--exit-on-failure` which exits immediately on any error and does not continue with any other charts in the course definition

If this is acceptable we could also shoot to add a summary of failures and reasons at the end of the run.

sudermanjr:
I would be in favor of defaulting to exit-on-failure and making --continue-on-failure an option, but that would be a larger departure from the existing UX.

endzyme:
I'm down with that too, I'll scope that as the change.

Here's what I've got so far:

- exit code 0 when everything runs successfully
- exit code 1 when either a chart or runhook fail
- immediately exit the app when any failures are encountered 
- add a `--continue-on-error` flag to override the immediate exit behavior

Thanks again everyone for the help identifying the need here.

willnewby:
> I'm down with that too, I'll scope that as the change.
> 
> Here's what I've got so far:
> 
>     * exit code 0 when everything runs successfully
> 
>     * exit code 1 when either a chart or runhook fail
> 
>     * immediately exit the app when any failures are encountered
> 
>     * add a `--continue-on-error` flag to override the immediate exit behavior
> 
> 
> Thanks again everyone for the help identifying the need here.

This sounds rad, and fits the pattern I would expect.
Thanks!

endzyme:
this has been released in `v1.4.0`

PR #88, opened 2019-04-11
MVP: Added End to End testing of Reckoner

endzyme:
- Added a full end-to-end test with installing charts with reckoner on python 2 & 3
- Added a basic yaml course to test with

### DIFF
--- a/end_to_end_testing/basic.yml
+++ b/end_to_end_testing/basic.yml
@@ -19,12 +19,17 @@ charts:
     repository:
       git: https://github.com/goharbor/harbor-helm.git
     version: master
-  redis:
+  redis-env:
+    chart: redis
+    namespace: redis-test-namespace
+    repository: stable
+    values:
+      test_environ_var: ${test_environ_var}
+  redis-hook:
+    chart: redis
     namespace: redis-test-namespace
     repository: stable
     hooks:
       pre_install:
         - whoami
       post_install: nonexistant command here
-    values:
-      test_environ_var: ${test_environ_var}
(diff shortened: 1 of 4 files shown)


endzyme:
This is an MVP and isn't efficient or easily developed upon. I think there needs to be more thought put into the "tests" themselves. The MVP aspect is to assure it's possible to run reckoner against a kubernetes cluster with helm. The output results testing from there is the hard part.

end_to_end_testing/basic.yml
```diff
@@ -0,0 +1,26 @@
...
+  incubator:
+    url: https://kubernetes-charts-incubator.storage.googleapis.com
+  reactiveops-stable:
+    url: https://charts.reactiveops.com/stable
+  reactiveops-incubator:
+    url: https://charts.reactiveops.com/incubator
+minimum_versions: #set minimum version requirements here
+  helm: 0.0.0
+  reckoner: 0.0.0
+charts:
```
sudermanjr: It would be super simple to add a couple hooks here that will add testing for that path. 

I know there's further expansion planned, but this would be a nice quick win.

endzyme: :+1:

endzyme: So I don't think reckoner returns a bad status code when hooks fail - it just skips the chart. I'll have to test to make sure the test would be valid - at least it would break even if it would cost some debugging :D

sudermanjr: I think that relates a bit to #86.  Definitely not a blocker for this PR.

endzyme: I did get an example running but it definitely isn't testing in enough isolation - MVP :D at least we've got something for a start. Still could have false positives and false negatives. We'll keep improving as we find bugs.

PR #91, opened 2019-04-12
Added dependency updates

endzyme:
* Updated GitPython, coloredlogs, and semver to use greater than or equal to latest


endzyme:
ready for testing

Issue #89, opened 2019-04-11
Documentation Needed


endzyme:
* Docs on releasing with new dynamic version process
* Docs on how to setup development environment with new requirements.txt
* Docs on differences between development-requirements.txt and the setup.py requirements

PR #92, opened 2019-04-15
Added docs for dev env and release for python3

endzyme:
#89

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,15 +4,17 @@
 ## Installation for Local Development
 
 Requirements
-* Python 3.7
-* pip3
-* VirtualEnv
+* Python 3
+* pip
 
+```shell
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -4,15 +4,17 @@
...
-* Python 3.7
-* pip3
-* VirtualEnv
+* Python 3
+* pip
 
+```shell
+%> python -V                   # Check your version of python
+%> python -m venv .venv        # setup a virtual env in your cloned repo
+%> source ./venv/bin/activate  # use the virtual env
```
sudermanjr: I think maybe this should be `source ./.venv/bin/activate` since lin 12 has `.venv` as the name.  This would make the resulting `(venv)` bits read `(.venv)`.

endzyme: derp! thanks :)

PR #93, opened 2019-04-24
Supporting PyInstaller (Binary Distribution of Reckoner)

endzyme:
I have incorporated changes that support the following:

* pip installs remain the same
* new support for pyinstaller compiling for binary distribution of reckoner
* refactored the end-to-end tests to use pyinstaller compiled binary
* refactored end-to-end tests to use VM with docker locally instead of docker-remote session

PyInstaller Notes:
* Binaries will only work on linux (today) and only support glibc based OS (not alpine)


installer/reckoner.spec
```diff
@@ -0,0 +1,32 @@
+# -*- mode: python -*-
+
+block_cipher = None
+
+
+a = Analysis(['cli.py'],
+             pathex=['.', '../', '/home/nhuanca/dev/reactiveops/reckoner/installer'],
```
sudermanjr: Should this have your path in it?  I think maybe we don't even need this file.

endzyme: haha neato :D must be autogenerated

PR #95, opened 2019-04-24
Adding darwin builds on tags

endzyme:
Missed one of the stanzas which builds os x when tags are pushed


endzyme:
#97 also fixes this - making it one commit so i can tag and push a release

PR #99, opened 2019-05-13
support helm wrapper plugins

nirdothan:
Helm wrapper style plugins such as [helm-secrets](https://github.com/futuresimple/helm-secrets) are invoked by adding the plugin name before the helm command .e.g `helm secrets install mychart -f /path/to/my-secrets.yaml`
A `plugin` item has been added to the `chart` dict. If provided in the plot, the plugin name is passed to the helm commend before in helm command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.7]
+- Support helm wrapper plugins such as [Helm Secrets](https://github.com/futuresimple/helm-secrets)
+
 ## [1.1.6]
 - Skipped versions to kickstart pipelines
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [1.1.7]
+## [1.2.0]
 - Support helm wrapper plugins such as [Helm Secrets](https://github.com/futuresimple/helm-secrets)
 
 ## [1.1.6]
--- a/README.md
+++ b/README.md
@@ -147,6 +147,20 @@ charts:
 
 The version of the chart to use
 
+### plugin
+
+A helm wrapper plugin to invoke for this chart.
+Make sure that the plugin is installed in your environment, and is working properly.
+Unfortunately, we cannot support every plugin.
+
+```
+charts:
+  chart_name:
+    plugin: secrets
+    files:
+      - /path/to/secret/values/file.yaml
+```
+
 ## Repository Options
 
 ### name
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -355,7 +355,14 @@ class Chart(object):
         """
         for key, value in self.values_strings.items():
             for k, v in self._format_set(key, value):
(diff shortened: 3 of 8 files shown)


endzyme:
Thank you Nir! We appreciate the PR! We'll be able to take a look this week sometime.

ejether:
This is good foundation for this: https://github.com/reactiveops/reckoner/issues/3

sudermanjr:
> This is good foundation for this: #3

I thought that too, until I looked into the [helm diff plugin](https://github.com/databus23/helm-diff) and it doesn't seem to support `helm upgrade --install`.  So it would take some reworking of our current workflow to use that particular plugin.

ejether:
Booo

endzyme:
@nirdothan - Thanks again for the contribution. We saw that you added some embedded logic to make different decisions for the secrets plugin vs the regular helm commands. I've submitted a patch the plugin provider (which you can use locally until it's merged) which will fix support for `-f` or `-f=` formatting. (https://github.com/futuresimple/helm-secrets/pull/118)

I also have some small patch changes (see [patch here](https://gist.github.com/endzyme/0a2aaf8459d5e1edaa0b17fc2278652b)) that would limit the changes to the helm command provder, and thus help us in our test coverage. I've also added a test to assure plugin support is tested for regressions.

Let me know if these changes make sense and if it still works as you'd expect. 

If you wanted to apply the patch file linked above to your PR you could do so with `git apply patchfile.patch` and then perform a commit and push as normal.

As always, feel free to ping us with any questions :) thanks again and hope we can get this merged in soon!

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.7]
```
sudermanjr: Could you please make this a minor version change `1.2.0` since this will add functionality to the tool?

nirdothan:
> I also have some small patch changes (see [patch here](https://gist.github.com/endzyme/0a2aaf8459d5e1edaa0b17fc2278652b)) that would limit the changes to the helm command provder, and thus help us in our test coverage. I've also added a test to assure plugin support is tested for regressions.

@endzyme I tested the patch and it seems to insert the plugin name in the wrong location in the helm command-line. See DEBUG log below. The correct location is after `helm` and before `upgrade`. Putting it as the first argument is too far.

```
DEBUG {'version': '0.2.1', 'plugin': 'secrets', 'files': ['/path/to/values/file.yml'], 'values': {}, 'set_values': {}, 'values_strings': {}}
DEBUG helm upgrade --recreate-pods --install secrets heapster stable/heapster --namespace=kube-system --kube-context=minikube --version=0.2.1 -f=/path/to/values/file.yml
```

endzyme:
Sorry about that @nirdothan - totally my bad - I've fixed it up - with much of your code in play. I added some tests to support regressions on this. https://gist.github.com/endzyme/0a2aaf8459d5e1edaa0b17fc2278652b

nirdothan:
https://github.com/reactiveops/reckoner/blob/2346e3dc268b5b9590924f76f7b799ba5686fb7f/reckoner/chart.py#L338-L345

I wonder if we should use this condition only for `secrets` plugin, for all plugins, or just refrain from using `option=value` style regardless. 
It seems to me that the helm [documentation](https://helm.sh/docs/using_helm/#the-format-and-limitations-of-set) doesn't even mention `option=value`, only `option value`.

endzyme:
That's an interesting find @nirdothan, thanks for digging that up. I think helm is using Cobra under the hood, which is why it supports both `var=value` and `var value`. I am running a local copy with a refactor to spaces from equal signs and we can investigate that.

Would it be acceptable to move forward with the plugin support using `=` separator and then we can open another PR to refactor the app to use spaces instead of equal signs?

PR #104, opened 2019-05-17
WIP Refactored the helm arguments to implement spaces

endzyme:
Some of the plugin ecosystems parse opts with bash and thus sometimes can get hung up on `-f=this.file` vs `-f this.file` -- with the error "Cannot find file: "=this.file".

The Popen module should handle all necessary escaping by default with anything passed into args so even setting `--set my_key=has spaces in value` should be handled correctly.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -229,9 +229,10 @@ jobs:
             reckoner plot --only nginx-ingress end_to_end_testing/basic.yml
             helm list --output json | jq -e '.Releases[]|select(.Name == "nginx-ingress")|.Status == "DEPLOYED"'
 
-            reckoner plot --only nginx-ingress --only redis-env end_to_end_testing/basic.yml
+            test_environ_var=somevalue reckoner plot --only nginx-ingress --only redis-env end_to_end_testing/basic.yml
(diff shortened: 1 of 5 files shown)


endzyme:
@nirdothan if you wanted to test this out locally - that would be awesome (to assure it satisfies the needs for your `secrets` plugin workflow)

nirdothan:
Will try next week

nirdothan:
We tested the branch today with helm secrets. Works fine.

endzyme:
Sounds great. I'll be running this locally for a week along with some others to see if there are any edge cases where this won't work. @nirdothan is it ok to release this on Friday?

endzyme:
From using this for a while now and I think we're ok to start merging and releasing. Thanks again for the contribution and the patience @nirdothan

nirdothan:
Thank you very much

endzyme:
I'm merging this today. Our other code owner is away but I'm pushing forward with a final review later on. I've been running this for a week and I know Andy has been as well. I've also had some other confirmations on basic tests with the end-to-end tests as well as with other folks at ReactiveOps. 

I'll merge and cut a release shortly.

endzyme:
Merging this outside of normal PR review - will revisit a retro review when the other code owner is back online :) :wave:

PR #108, opened 2019-06-11
adding .vscode to .gitignore

bambash:


bambash:
added ignore to global gitconfig

PR #110, opened 2019-06-25
Deprecating Python2 and requiring ~=3.6

sudermanjr:
This removes the ability to install Reckoner in a python2 environment. There are notes in both the README and CONTRIBUTING docs, as well as disabling python2 in the setup.py.  CI for python2 has also been removed.

Example output when trying to pip install in python2:

```
 reactiveops/reckoner   sudermanjr/deprecate-python-2 ✘   .venv-2 
└─ pip install .
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7.
Processing /Users/asuderma/repos/reactiveops/reckoner
ERROR: reckoner requires Python '~=3.6' but the running Python is 2.7.15
```

Setup.py changes based on docs here: https://packaging.python.org/guides/distributing-packages-using-setuptools/

Also in this PR:
* Adding codecov badge.
* Removed call to codecov on python 3.6 build.  Only need to run codecov
script once.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,9 +15,6 @@
 version: 2.1
 
 executors:
-  python-2-7:
-    docker:
-      - image: circleci/python:2.7
   python-3-6:
     docker:
       - image: circleci/python:3.6
@@ -32,22 +29,6 @@ executors:
       xcode: 10.2.0
 
 jobs:
-  build-2-7:
-    executor: python-2-7
-    working_directory: ~/reckoner
-    steps:
-      - run:
-          name: Setup PATH to support pip user installs
-          command: echo 'export PATH=$PATH:/home/circleci/.local/bin' >> $BASH_ENV
-      - checkout
-      - run:
-          name: Unit Tests
-          command: |
-            pip install --user -r development-requirements.txt
-            pip install --user -e .
-            reckoner --version
-            pytest --cov ./
-            codecov
   build-3-6:
     executor: python-3-6
     working_directory: ~/reckoner
@@ -62,8 +43,7 @@ jobs:
             pip install --user -r development-requirements.txt
             pip install --user -e .
             reckoner --version
-            pytest --cov ./
-            codecov
+            pytest
   build-3-7:
     executor: python-3-7
     working_directory: ~/reckoner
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -52,12 +53,6 @@ For development see [CONTRIBUTING.md](./CONTRIBUTING.md).
   * `generate`: Generates example file `course.yml` with extensive descriptions
   * `version`: Output reckoner version
 
-## Example configuration file:
```
endzyme: nice catch

endzyme:
This looks extremely awesome - I think the only other consideration is to mark a new version for this PR, and say that going forward from `v1.3.0` there won't be any support for Python2 (something like that would be good in the changelog).

README.md
```diff
@@ -1,5 +1,4 @@
-
-# reckoner [![CircleCI](https://circleci.com/gh/reactiveops/reckoner.svg?style=svg)](https://circleci.com/gh/reactiveops/reckoner)
+# reckoner [![CircleCI](https://circleci.com/gh/reactiveops/reckoner.svg?style=svg)](https://circleci.com/gh/reactiveops/reckoner) [![codecov](https://codecov.io/gh/reactiveops/reckoner/branch/master/graph/badge.svg)](https://codecov.io/gh/reactiveops/reckoner)
```
endzyme: This might be a holdover from the other PR? Not sure if it's supposed to be in here

sudermanjr: I think maybe I forgot it, or maybe it is a holdover.  Either way I'll take care of it when I rebase.

sudermanjr: I must have forgotten to add the codecov badge in the other PR.  I added it to the description on this one as well as removing the duplicate call to codecov.

.circleci/config.yml
```diff
@@ -62,8 +43,7 @@ jobs:
             pip install --user -r development-requirements.txt
             pip install --user -e .
             reckoner --version
-            pytest --cov ./
-            codecov
+            pytest
```
endzyme: :100:

PR #111, opened 2019-06-26
Adjustments for end to end tests

endzyme:
- Refactored all end to end tests into a file
- Added specific end to end course files to test specific behavior in isolation
- Added structure for future tests and output to show which tests fail
- Feature Added: ALL tests will run every time, not just fail on the first broken test
- Added light "framework" for adding tests and reporting errors


endzyme:
Downside to this change are:
- The end to end tests now take an additional minute or so to run
- Little more complexity in writing end to end tests
- More yaml files for course tests

Upsides:
- Better isolation in end to end testing of features
- An easier way to read/write tests
- Allows us to test exit codes now

end_to_end_testing/setup.sh
```diff
@@ -0,0 +1,43 @@
...
+
+echo "Installing git and jq"
+sudo apt-get install -yqq jq git
+
+echo "Installing Reckoner"
+sudo cp /tmp/binaries/reckoner-linux-amd64 /usr/local/bin/reckoner
+reckoner version
+
+echo "Installing Kind"
+curl -sLO https://github.com/kubernetes-sigs/kind/releases/download/0.2.1/kind-linux-amd64
```
sudermanjr: Should we update this to 0.4.0 now?  Also, should we pin the Kubernetes version we are using for this?  I think 0.4.0 now defaults to 1.14 or something like that.

endzyme: I will adjust that in a new PR if that's ok - good call out and I will make sure to Pin both the Kind version to 0.4.0 and K8s version to v1.14.x

bambash:
+1 on pinning the kubernetes and kind versions

:+1: looks great!

Issue #101, opened 2019-05-13
--local-development flag does not work


sudermanjr:
The `--local-development` flag suggests that it does not need Tiller access to work, however when using it with `--dry-run` I see this error:

```
2019-05-13 14:41:27 root[16325]    ERROR Command Failed with output below:
STDOUT:
STDERR: Error: Get http://localhost:8080/api/v1/namespaces/kube-system/pods?labelSelector=app%3Dhelm%2Cname%3Dtiller: dial tcp [::1]:8080: connect: connection refused

COMMAND: helm upgrade --install metrics-server stable/metrics-server --namespace=metrics-server --dry-run --debug --version=2.4.0 --set=args[0]=--kubelet-insecure-tls --set=args[1]=--kubelet-preferred-address-types=InternalIP --set=args[2]=--metric-resolution=30s --set=args[3]=--v=3
```

endzyme:
I think I need to get the context of original intent for `--local-development`. The README sounds like it's a flag for people developing on Reckoner itself to be able to run through logic without running real commands on helm. If that's the case then I think maybe we should evaluate if it's a commonly used feature. 

If it's to test your course file or provide some other value, then maybe the flag should be a validate type flag instead. 

Any insight you may have @ejether or @sudermanjr would be awesome if you've got some cycles.

ejether:
The flag was there so that I could hack on it when is on an airplane and my minikube had broken and I didn't have a cluster to work from. I wanted it to, essentially, output the generated helm command.
It was intended to not need access to tiller but after the refactor and as the code path got a little more complicated, wrapping so many different aspects in differnet methods got complicated and things were missed. 

I don't have a strong opinion any more if this is a useful flag to maintain but if you look at it as a slightly broken, or leaky feature flag, it might make clarify it's intent.

sudermanjr:
Once helm3 becomes default, this will be a thing of the past.   I vote we just leave it the way it is or remove it.

endzyme:
After some discussion today, I think we'll use this issue to pull out the flag. Thanks again for all your help on with this info

endzyme:
resolved in #112 and #113

PR #112, opened 2019-07-02
Adjustments for removing local-development flag

endzyme:
Changes:
- Removed all references in code and tests for `local_development` or `--local-development` flag
- Adjusted tests to have to incorporate logic for missing flag

The `--local-development` flag was used in tests to sometimes "not call" mocks or certain code areas. With this missing, we need to have better coverage and splits on the logic that this gate was hiding (future work). The tests now all pass and are functioning as expected. `tests/test_reckoner.py` still remains brittle and we could possible get rid of this test file in favor of end to end testing. Lots of the tests in `./tests/test_reckoner.py` are halfway between integration and end-to-end tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,10 +17,10 @@ version: 2.1
 executors:
   python-3-6:
     docker:
-      - image: circleci/python:3.6
+      - image: circleci/python:3.6-stretch
   python-3-7:
     docker:
-      - image: circleci/python:3.7
+      - image: circleci/python:3.7-stretch
   vm:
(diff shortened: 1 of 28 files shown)


endzyme:
#101

endzyme:
Code coverage went "down on the diff" due to the code that was moved not being covered (wasn't originally either, just in a new line position with the changes I made). 

Some of the files also increased or decreased coverage due to the same reasons

endzyme:
Moving this forward since it's already incorporated in #113

PR #113, opened 2019-07-02
Exit codes for exit status

endzyme:
For Issue #86 

    Added logic:

    - Exit codes
    - Exit after first error
    - bubble up the issue at the end of the message

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,7 +164,7 @@ jobs:
       - run:
           name: find release tag
           command: |
-            release_id=$(curl -s -H "Authorization: token ${GITHUB_ACCESS_TOKEN}" -X GET https://api.github.com/repos/reactiveops/reckoner/releases/tags/${CIRCLE_TAG} | jq -e .id)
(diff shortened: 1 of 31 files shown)


endzyme:
This change requires #112 and #111

sudermanjr:
Ran a couple quick tests with this, and all the intended functionality seems to work really well.  Super exciting stuff and really awesome.  

Is the `continue-on-error` flag documented anywhere besides the help command?  Also, I think we might need to add a mention of these features to the changelog

endzyme:
> Ran a couple quick tests with this, and all the intended functionality seems to work really well. Super exciting stuff and really awesome.
> 
> Is the `continue-on-error` flag documented anywhere besides the help command? Also, I think we might need to add a mention of these features to the changelog

Good catch - i'll publish changes in changelog and make sure it's mentioned in the Readme. I've opened #117 to help clean up docs holistically, but for now I'll add to README.md.

endzyme:
@bambash @sudermanjr one last looksy before I merge this on through if you've got some time soon :) thanks again for all the testing

PR #121, opened 2019-07-16
add roadmap, quickstart

bobby-brennan:
Quickstart example isn't currently working. Grafana installs, but Polaris fails.


bobby-brennan:
Thanks Nate! With some help from Nick I modified it a bit to be more explicit on how to set the chart name:

```
charts:
  polaris-dashboard:
    namespace: polaris-dashboard
    repository:
      git: https://github.com/FairwindsOps/charts
      path: stable
    chart: polaris
```

PR #127, opened 2019-07-25
Rename ReactiveOps to Fairwinds

bobby-brennan:
Remaining mentions: 
```

```

### DIFF
--- a/reckoner/meta.py
+++ b/reckoner/meta.py
@@ -2,7 +2,7 @@
 
 # pylint: skip-file
 
-# Copyright 2019 ReactiveOps Inc
+# Copyright 2019 FairwindsOps Inc
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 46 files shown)


reckoner/meta.py
```diff
@@ -27,4 +27,4 @@
     from pkgutil import get_data
     _raw_ver = get_data(__distribution_name__, 'version.txt').decode('UTF-8', 'ignore').rstrip("\r\n")
     __version__ = re.sub(__version_modifier__, r'\g<1>-\g<2>', _raw_ver)
-__author__ = 'ReactiveOps, Inc.'
+__author__ = 'Fairwinds, Inc.'
```
endzyme: We're copy writing to `FairwindsOps Inc` and authoring to `Fairwinds, Inc.` -- I actually don't know what is the best one to use here (if there are legal implications)

endzyme:
lookin sharp :+1: (other than the name question)

reckoner/meta.py
bobby-brennan: Good catch! The comma screwed up my regex :) fixed.

PR #131, opened 2019-08-06
Added strong type support for course.yml

endzyme:
Added the ability to keep types strong by using the `values: {}` section of a chart config. Now the `set-values` and `values` blocks have different behavior underlying.

**Breaking Changes**
The `chart: {values: {}}` config block and `chart: {set-values:{}}` config block now have different behavior. `set-values: {}` always gets translated into `helm` arguments as `--set key=value`. The `values: {}` now gets applied to `helm` arguments as `-f temporary_values_name.yml`. The `values: {}` config block is now fully consistent with intended types and would behave as though you are using a `-f my_values.yml` in your `helm` command. Prior to this change you would see inconsistent type casting for float, bool and integer config settings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [2.0.0]
+### Breaking Changes
+- Changes to `values: {}` behavior:  
+  The chart: {values: {}} config block and chart: {set-values:{}} config block now have different behavior. set-values: {} always gets translated into helm arguments as --set key=value. The values: {} now gets applied to helm arguments as -f temporary_values_name.yml. The values: {} config block is now fully consistent with intended types and would behave as though you are using a -f my_values.yml in your helm command. Prior to this change you would see inconsistent type casting for float, bool and integer config settings. For more information on the behavior differences between `values`, `set-values`, and `values-strings` you can look at our [end to end testing test course.yml](/end_to_end_testing/test_strong_typing.yml).
+- BUG FIX: Using `"null"`, `null`, `"Null"`, and `"NULL"` as values in `set-values: {}` will be interpreted as `null` (void of value) in the `--set` value. Previously, if you set `null` as the yaml value, you would get `--set key=None` due to python interpreting the value as `None` and thus would show up as `{"key": "None"}` in the helm values. This fix more closely aligns with expected behavior in helm.
+
 ## [1.4.0]
 ### Breaking Changes
 - Removed the `--local-development` flag from `plot` command (unused) and cleaned up test dependencies
--- a/end_to_end_testing/run_end_to_end.sh
+++ b/end_to_end_testing/run_end_to_end.sh
@@ -14,17 +14,31 @@ set -o errtrace
 
 E2E_FAILED_TESTS=false
 E2E_FAILED_MESSAGES=()
+E2E_SKIPPED_MESSAGES=()
+
+function print_status_end_exit() {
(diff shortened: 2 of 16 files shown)


end_to_end_testing/run_end_to_end.sh
```diff
@@ -57,7 +85,30 @@ function helm_release_has_key_value() {
...
+    local key="${2}"
+
+    if helm get values "${release_name}" --output json | jq -e ". | has(\"${key}\")" &>/dev/null; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+function helm_release_key_value_is_type() {
```
bobby-brennan: It looks like this function is being called inside a `for` loop, resulting in many separate, identical calls to `helm get values`. Would it make more sense to grab that output once and then pass that to `value_is_type` and `values_has_key`?

Edit: realizing this is just a test file, so NBD

end_to_end_testing/run_end_to_end.sh
```diff
@@ -14,24 +14,52 @@ set -o errtrace
...
+trap 'echo Failed unexpectedly on line ${BASH_LINENO} running ${FUNCNAME[0]}: ${BASH_COMMAND}; clean_helm; exit 1' ERR
 
 # Change to the script dir for help finding yamls
 cd "$(dirname "${0}")"
 
+# Helper to clean out all the stuff between tests
+function clean_helm() {
+    if [ -z "${E2E_LEAVE_INSTALLED_CHARTS}" ]; then
+        # Get all installed things in helm and delete/purge them
+        helm list --output json | jq '.Releases[].Name' | xargs -I {} helm delete --purge {}
```
sudermanjr: Do we publish a minimum helm version for Reckoner?  I was just thinking that --output json didn't show up until like 2.13.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.5.0]
+### Breaking Changes
```
bobby-brennan: Should the README be updated to use `values` rather than `set-values` in the Quickstart?

Also, should `values-strings` be deprecated now, since you'd just have to explicitly make something a string in your yaml? like
```yaml
values:
  bar: "2"
```

endzyme: Actually, values-strings is exposing a feature in helm where you can set `--set-string` (see [Docs Here](https://helm.sh/docs/helm/#options-43))

end_to_end_testing/run_end_to_end.sh
endzyme: ah-ha - interesting. We don't actually publish and support matrix, which we could do but I think it should probably be a real effort to perform continual testing. Maybe we adhere to helm stable + previous minor version?

end_to_end_testing/run_end_to_end.sh
endzyme: We could definitely rely on a cached version of the objects captured after each reckoner plot command, although bash has it's limitations with variable size and string manipulation. Generally when stuff like this gets involved, it means we should probably move away from bash. I'm happy to chat about this more, but if this isn't blocking the PR then we maybe could open an issue.

CHANGELOG.md
endzyme: Oops - didn't address your first question - yes the readme needs updating - i'll do that now. :) good catch

end_to_end_testing/run_end_to_end.sh
bobby-brennan: Given that it's just a test, I'd say leave it. I originally thought this was running on (potentially large) user inputs.

end_to_end_testing/run_end_to_end.sh
sudermanjr: Seems good to me.

endzyme:
FYI @ejether pointed out a bug that this isn't interpolating variables in the new temp yaml values files. A push has been made and should fix the issue. Also added end-to-end test coverage for this logic.

PR #133, opened 2019-08-09
Added Helm 3 detection and warning

bambash:
Added Helm 3 detection to `HelmClient`. Issues a warning to the user that Helm 3 is in use and they should proceed with caution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.5.0]
+- Added Helm 3 detection
+
 ## [1.4.0]
 ### Breaking Changes
 - Removed the `--local-development` flag from `plot` command (unused) and cleaned up test dependencies
--- a/reckoner/helm/client.py
+++ b/reckoner/helm/client.py
@@ -71,11 +71,11 @@ class HelmClient(object):
 
     @property
     def client_version(self):
-        return self._get_version('--client')
+        return self._get_version(kind='--client')
 
     @property
     def server_version(self):
-        return self._get_version('--server')
+        return self._get_version(kind='--server')
 
     @property
     def repositories(self):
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.5.0]
```
endzyme: This will have a collision - let's avoid updating changelog for now - I think this project should probably be run as a parallel branch to master that we can later tune, or merge in after we're all done with the POC.

endzyme: I've made a new branch called `helm-3-work` so we can do reviews and merge into this branch. This will help us run in parallel and not have to worry too much about changelog for now.

reckoner/helm/client.py
```diff
@@ -172,7 +179,12 @@ def _find_version(raw_version):
         if ver:
             return ver.group(1)
         else:
-            return None
+            logging.debug("Couldn't find Helm version. Are we using Helm 3?")
+            try:
+                ver = re.compile(r'v([0-9\.]+)(\+[a-zA-Z]+)(\+g[0-9a-f]+)?').search(str(raw_version))
```
endzyme: Having compiles within a method can be inefficient for memory since you're continually initializing it.

Note this is not a problem for performance but just a heads up as to why the original implementation is a class variable

Example of performance:
```python
import re, timeit

my_regexp = re.compile(r'a')

def outside_of_func():
        my_regexp.search(str("asdf"))

def inside_of_func():
        r = re.compile(r'a').search(str("asdf"))

if __name__ == '__main__':
        print("inside")
        print(timeit.timeit(lambda: inside_of_func(), number=100000))
        print("outside")
        print(timeit.timeit(lambda: outside_of_func(), number=100000))
```

reckoner/helm/client.py
```diff
@@ -152,9 +152,16 @@ def _clean_non_global_flags(list_of_args):
...
+    def _get_version(self, kind):
+        try:
+            get_ver = self.execute("version", arguments=['--short', kind], filter_non_global_flags=True)
+            ver = self._find_version(get_ver.stdout)
+        except HelmClientException as e:
+            # check if running helm3, --client and --server flags are removed
+            logging.debug("Couldn't find Helm version. Are we using Helm 3?")
+            get_ver = self.execute("version", arguments=['--short'], filter_non_global_flags=True)
+            ver = self._find_version(get_ver.stdout)
+            logging.warn("🔥️ 🐲️  Detected Helm 3 as version in use. Helm 3 is untested and not supported. Proceed with caution! 🔥️ 🐲️")
```
sudermanjr: Icons are awesome, but how do they work with a screen reader?

sudermanjr:
Looks like the tests are failing due to this change.

endzyme:
I am going to close this for now - to be reopened later - Nate has been away but this is currently a work in progress (spoke with him a little while ago about it in chat)

PR #134, opened 2019-08-19
Nb/helm3 detection

bambash:
Added Helm 3 detection to HelmClient. Raises an exception that Helm 3 is in use and not currently supported.

### DIFF
--- a/reckoner/helm/client.py
+++ b/reckoner/helm/client.py
@@ -71,11 +71,11 @@ class HelmClient(object):
 
     @property
     def client_version(self):
-        return self._get_version('--client')
+        return self._get_version(kind='--client')
 
     @property
     def server_version(self):
-        return self._get_version('--server')
+        return self._get_version(kind='--server')
 
     @property
     def repositories(self):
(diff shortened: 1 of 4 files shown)


reckoner/helm/client.py
```diff
@@ -152,10 +153,20 @@ def _clean_non_global_flags(list_of_args):
...
+        try:
+            get_ver = self.execute("version", arguments=['--short', kind], filter_non_global_flags=True)
+            ver = self._find_version(get_ver.stdout)
+        except HelmClientException:
+            # check if running helm3, --client and --server flags are removed
+            logging.debug("Caught exception when checking version. Are we using Helm 3?")
+            get_ver = self.execute("version", arguments=['--short'], filter_non_global_flags=True)
+            ver = self._find_version(get_ver.stdout)
+            if ver.startswith('3'):
+                logging.error("\n\n🔥️ 🐲️  Helm 3 is untested and not supported. 🔥️ 🐲️\nhttps://github.com/FairwindsOps/reckoner/issues/118")
```
sudermanjr: Emojis are cool, but how do they behave with a screenreader?

sudermanjr:
The end-to-end tests are failing on package install.  Might require an `apt-get update` to be run before the apt-get install perhaps?

sudermanjr:
I opened https://github.com/FairwindsOps/reckoner/pull/135 which will fix the end-to-end tests.

PR #136, opened 2019-08-27
Supporting Schema Validation

endzyme:
Phase 1: Support validating the schema so that users
         get notified when there is an extra param or
         if they have an incorrect key/value pair in
         their course.yml.

I have housed the logic in a new module that reads the yaml and also
returns the correct errors expected at the cli.py. Earlier I was
planning on integrating this into course.py, as there's where the
initial YAML read happens, but since it's a part of __init__ in that
class, it makes loads of extra work for mocking. I opted to move this
ealier in the flow and have it fail faster even if we have to read the
course.yml twice. It's now integrated as the first thing in the cli.py.


endzyme:
@rbren I think we're ready for a review - i need some help doing some exploratory testing with pip install method on python 3 too if anyone has time.

endzyme:
@bambash @sudermanjr Do you all have any time this or next week to review?

sudermanjr:
> @bambash @sudermanjr Do you all have any time this or next week to review?

I'll take a look this week.

bambash:
Tested this out today, works great - awesome addition

sudermanjr:
I found one potential issue with this change.  

In the past, there has been a pattern of creating reference yaml blocks that can be re-used later.  For example at the top of a file:

```
ingress-common-values: &ingress-common-values
  controller.service.externalTrafficPolicy: "Local"
  controller.publishService.enabled: "true"
  defaultBackend.name: "default-backend"
  rbac.create: "true"
  controller.service.targetPorts.https: 443
  controller.service.targetPorts.http: 80
  controller.metrics.enabled: true
  controller.stats.enabled: true
  controller.extraArgs.default-ssl-certificate: "kube-system/wildcard"
  controller.livenessProbe.initialDelaySeconds: 120
ingress-common-values-strings: &ingress-common-values-strings
  controller.podAnnotations:
    ad\.datadoghq\.com/nginx-ingress-controller\.check_names: |
      ["prometheus"]
    ad\.datadoghq\.com/nginx-ingress-controller\.init_configs: |
      [{}]
    ad\.datadoghq\.com/nginx-ingress-controller\.instances: |
      [
        {
          "prometheus_url": "http://%%host%%:10254/metrics"\,
          "namespace": "ingress"\,
          "metrics": ["nginx*"\, "ingress*"]\,
          "max_returned_metrics": "18000"
        }
      ]

```

Maybe we could add a `references` to the schema and then put these under that key?

endzyme:
@sudermanjr RE: the "reusable bits" in YAML - would it be acceptable to place those into a key like `_references` or something that can be included in the schema?

endzyme:
This PR is also missing docs on the README about schema validation and what a valid schema looks like

ejether:
That will work from a yaml standpoint but would require a change in many files.

sudermanjr:
> @sudermanjr RE: the "reusable bits" in YAML - would it be acceptable to place those into a key like `_references` or something that can be included in the schema?

Definitely.  I think I mentioned that possibility in my comment ;-)

sudermanjr:
> That will work from a yaml standpoint but would require a change in many files.

Are you referring the the addition of a `references` to the schema?

I think that is a reasonable ask of the users considering the benefits to be had from this schema validation.  What do you think?

ejether:
I was referring to the centralization of the anchors into a `_references` key would be a lot of effort. It would probably be worth it but is there a way to validate the scheme without forcing that effort by evaluating the references before schema validation? `yaml -> dict -> yaml -> validate`

Caveat: That may be a stupid suggestion and I've not been involved in writing the code so if that idea has already been thrown out, ignore me.

endzyme:
> I was referring to the centralization of the anchors into a `_references` key would be a lot of effort. It would probably be worth it but is there a way to validate the scheme without forcing that effort by evaluating the references before schema validation? `yaml -> dict -> yaml -> validate`
> 
> Caveat: That may be a stupid suggestion and I've not been involved in writing the code so if that idea has already been thrown out, ignore me.

To @ejether: The yaml will be derefenced (i.e. rendered to the verbose view with duplicates etc) before the schema validation happens. Duplicate key checking will happen before schema validation but that's the only check (outside of just syntactical checks). 

To @sudermanjr: OOPS yeah I definitely just read something, and subconsciously replayed it (ha!) - so yes, I think that putting things into a defined key, and then having the schema just accept any values below there (free form) would be best. 

Here's how I'm envisioning this change working:

1. Communicate that we're enabling schema validation and have a doc in the repo that explains what this means to previous versions
2. Anyone with reference links in YAML (the `&thing` to `*thing` notation) will have to put those blocks into a `_references` block (so it can be a standard part of the schema).
3. When reading a course with `_references` the code would
    a. Read the YAML into an ordered dictionary (ruamel.yaml default) and fail to load yaml if there are duplicate YAML keys, or if there are YAML syntax errors
    b. Run the serialized data into the schema checker and fail on schema validations
    c. continue with the normal code path

Sound ok?

ejether:
Re: 2.
I still think that _requiring_ all references to be under a new key or validation breaks is an unnecessary pain point for the user when we don't need to. Making it hard to upgrade means people are less likely to. 

Also, in doing some more reading, it seems like the canonical way for declaring anchors and aliases is inline so that it is declared inline the first time and aliased from there (as opposed to a single block of anchors) but I'm less sure of that. 

Since this affects everyone at FW I suggest bring it up at Office Hours and seeking confirmation from the users, as long as the team is on board with the plan.

endzyme:
> Re: 2.
> I still think that _requiring_ all references to be under a new key or validation breaks is an unnecessary pain point for the user when we don't need to. Making it hard to upgrade means people are less likely to.
> 
> Also, in doing some more reading, it seems like the canonical way for declaring anchors and aliases is inline so that it is declared inline the first time and aliased from there (as opposed to a single block of anchors) but I'm less sure of that.
> 
> Since this affects everyone at FW I suggest bring it up at Office Hours and seeking confirmation from the users, as long as the team is on board with the plan.

Thanks for the insight. I agree that the least amount of "overhead" in upgrading would be best. We haven't really discussed the inverse implications of allowing any old key at the top level. 

Original intent of schema validation:
1. Help people avoid defining unused and incorrectly set yaml for their course files
2. Provide a schema so people know what you can set for the course

The reason I had originally decided to only allow certain keys is so that the first intent listed above could be met. If we allow any key in the main root of the yaml (so you can set references at the top level) then there's no way to let a user know that they have an incorrect schema at the root level (typo in `namespace` or other examples like `min_version` vs `minimum_version`). If we were to put all reference type info into a single block, then we could maintain the original intent at the top level root of the course yaml. Here's an example of what we were envisioning above:

```yaml
_references:
  some_reusable_values: &reuse
    settings: "yes_setting"
    more_key: ""ore Value!
namespace: "infra"
charts:
  my-chart:
    namespace: "test"
    chart: "nginx-ingress"
    set-values: *reuse
    values:
      controller.name: "some-class"
```

I will add this to the agenda in engineering office hours to discuss but I'd like to have both sides of the coin to discuss further. The main discussion I hope will revolve around whether the overhead is worth the value.

endzyme:
Office hours decided we should go with `_references` block at the top level as a specific part of the schema. Just pushed that as well as better documentation.

Issue #145, opened 2019-10-14
Error when using `files` key


trankchung:
I'm trying to deploy a simple chart using the `files` key under a chart and it's throwing error. If i removed the `files` key it deployed fine.

*course.yaml*
```
---
namespace: kube-system

repositories:
  stable:
    url: https://kubernetes-charts.storage.googleapis.com

repository: stable

charts:
  nfs-client-provisioner:
    namespace: nfs-client-provisioner
    chart: nfs-client-provisioner
    version: 1.2.6
    files:
      - nfs-client-provisioner.yaml
```

*error*
```
$ ./reckoner plot sources.yaml
2019-10-14 15:29:01 my-desktop root[23061] INFO Installing nfs-client-provisioner
2019-10-14 15:29:01 my-desktop root[23061] ERROR Helm upgrade failed on nfs-client-provisioner
2019-10-14 15:29:01 my-desktop root[23061] ERROR ERROR: Chart failed to install.
2019-10-14 15:29:01 my-desktop root[23061] ERROR  - nfs-client-provisioner
2019-10-14 15:29:01 my-desktop root[23061] ERROR Stopping chart installations due to an error! Some of your charts may not have been installed!
⛵🔥 Encountered errors while running the course ⛵🔥

* * * * *

Chart Name: nfs-client-provisioner
Status: Failed
Error Reason: unsupported operand type(s) for +: 'NoneType' and 'str'
```

bambash:
Hi @trankchung, thanks for reporting this. Which version of reckoner are you encountering this on?

trankchung:
```
$ ./reckoner version
2.1.0```

I tried to convert it into `values` and I still get an error. FYI this is the first time I'm using Reckoner. The error for when I use `values` is below. The file `/tmp/tmph0gzjpme.yml` isn't there.

```COMMAND: helm upgrade --install nfs-client-provisioner stable/nfs-client-provisioner --namespace nfs-client-provisioner --kube-context test --dry-run --debug --version 1.2.6 -f /tmp/tmph0gzjpme.yml
```

bambash:
I was able to reproduce the `files` issue, but `values` is working as expected. Can you share a sample of how you're declaring `values:`?

We'll take a look at why `files` would be failing.

trankchung:
So I did several more tests and it's not `values` that causing the second error. It seems to be the name of the release is too long I think. When I use this it caused that error.
```
charts:
  nfs-client-provisioner:
    ...
```

[...]

PR #148, opened 2019-10-16
Fix values file pathing

bambash:
Relevant issue: https://github.com/FairwindsOps/reckoner/issues/145

This PR makes `course_base_directory` an absolute path. Added `os.path.join` to `chart.build_files_list()`


bambash:
Pushed a new test that ups code coverage a bit

PR #161, opened 2019-12-03
Update HELM3_BLOCKERS.md

yanivpaz1973:
Add version error handling


sudermanjr:
@yanivpaz1973 Thanks for the PR! Would you mind filing an issue for this instead? I think this is a bug in the way we handle the versions right now. This is supposed to work.

ejether:
this is fixed in :  https://github.com/FairwindsOps/reckoner/pull/162

PR #162, opened 2019-12-14
Ejether/update exception for helm 3 client

ejether:
This fixes an issue where when helm3 is installed, the helm2 client version check would fail and prevent the helm3 client version check from ever being run.  The combination of `HelmClientException` and `HelmVersionException` in the `except` will make it so it proceeds.

### DIFF
--- a/end_to_end_testing/setup_helm2.sh
+++ b/end_to_end_testing/setup_helm2.sh
@@ -1,0 +1,12 @@
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
+
+echo "Installing Helm"
+curl -sL https://storage.googleapis.com/kubernetes-helm/helm-v2.14.2-linux-amd64.tar.gz | tar xzv linux-amd64/helm
+sudo mv linux-amd64/helm /usr/local/bin/helm
+rm -rf linux-amd64
+helm version --client
+
+echo "Setting up Helm on Kubernetes"
+kubectl -n kube-system create serviceaccount tiller
+kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount kube-system:tiller
+helm init --service-account tiller --wait
--- a/end_to_end_testing/setup_helm3.sh
+++ b/end_to_end_testing/setup_helm3.sh
@@ -1,0 +1,17 @@
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
+
+echo "Installing Helm"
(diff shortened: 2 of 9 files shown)


end_to_end_testing/test_strong_typing.yml
```diff
@@ -3,6 +3,9 @@ namespace: testing #namespace to install the chart in, defaults to 'kube-system'
 minimum_versions: #set minimum version requirements here
   helm: 0.0.0
   reckoner: 0.0.0
+repositories:
```
ejether: This is probably unnecessary but isn't hurting anything.

end_to_end_testing/setup_helm3.sh
```diff
@@ -0,0 +1,17 @@
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
+
+echo "Installing Helm"
+curl -sL https://get.helm.sh/helm-v3.0.1-linux-amd64.tar.gz | tar xzv linux-amd64/helm
+sudo mv linux-amd64/helm /usr/local/bin/helm
+rm -rf linux-amd64
+helm version
+
+kubectl create namespace infra
```
sudermanjr: Should we mark this as a hack or a TODO since full Helm3 support for reckoner will probably be creating namespaces for us?

end_to_end_testing/setup_helm3.sh
```diff
@@ -0,0 +1,17 @@
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
+
+echo "Installing Helm"
+curl -sL https://get.helm.sh/helm-v3.0.1-linux-amd64.tar.gz | tar xzv linux-amd64/helm
```
sudermanjr: 3.0.2 dropped today, not sure if you want to update this.

end_to_end_testing/setup_helm2.sh
```diff
@@ -0,0 +1,12 @@
+source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
+
+echo "Installing Helm"
+curl -sL https://storage.googleapis.com/kubernetes-helm/helm-v2.14.2-linux-amd64.tar.gz | tar xzv linux-amd64/helm
```
sudermanjr: This version is also out of date. Probably need a better way to keep track of these for the future. Or run multiple version tests by parameterizing this  and exposing the parameter to Circle config.

ejether: Saw that. Elected not to tackle it at this time.

PR #174, opened 2020-02-03
Fixes #173. Regex the comments out before interpolating env vars.

### LINKED ISSUE #173: attempting to parse env vars inside comments
Looks like we may have a regression. 

https://github.com/FairwindsOps/reckoner/issues/150

sudermanjr:
- update to kind 0.7.0 for e2e tests
- update kube to 1.15.7
- add e2e and unit testing coverage for deeply-nested comments inside of `values`

### DIFF
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -16,6 +16,7 @@
 
 import logging
 import os
+import re
 
 from .kube import create_namespace, list_namespace_names
 from tempfile import NamedTemporaryFile as tempfile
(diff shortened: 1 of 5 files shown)


reckoner/chart.py
```diff
@@ -409,8 +407,11 @@ def build_files_list(self):
 
     @staticmethod
     def _interpolate_env_vars_from_string(original_string: str) -> str:
+
+        # We should never interpolate an env var in a comment. Strip comments from the string before interpolating.
+        comments_removed = re.sub(r'([^#]*)(#.*)', r"\g<1>", original_string)
```
ejether: I like this way better than the other old way.

ejether: nit. quote style? Intentional or meaningful?

sudermanjr: I think I was trying to get it to work and accidentally ended up with different styles. Will fix

PR #177, opened 2020-02-11
functional namespace management work with schema validation

ejether:
I ended up going with a schema that looks like:
In the course
```
namespace_managment:
  default:
    metadata:
      annotations:
        key:value
      labels:
        key:value
    settings:
      overwrite: true
```
and in the chart itself if desired
```
namespace_managment:
  metadata:
    annotations:
      key:value
    labels:
      key:value
  settings:
    overwrite: true
```


ejether:
@sudermanjr  @bambash Ready for review now.

sudermanjr:
> @sudermanjr @bambash Ready for review now.

I haven't forgotten, I promise. Planning to fully review this on Friday.

reckoner/kube.py
```diff
@@ -1,47 +1,146 @@
...
+        return self._metadata
+
+    @property
+    def overwrite(self) -> bool:
+        """ List of metadata settings parsed from the
+        from the chart and course """
+        return self._overwrite
+
+    def __load_config(self):
+        """ Protected method do load kubernetes config"""
```
sudermanjr:
```suggestion
        """ Protected method to load kubernetes config"""
```

reckoner/kube.py
```diff
@@ -1,47 +1,146 @@
...
+        try:
+            config.load_kube_config()
+            self.v1client = client.CoreV1Api()
+        except Exception as e:
+            logging.error('Unable to load kuberentes configuration')
```
sudermanjr:
```suggestion
            logging.error('Unable to load kubernetes configuration')
```

reckoner/chart.py
```diff
@@ -243,41 +250,41 @@ def update_dependencies(self):
...
+        - default_namespace (string). Passed in but will be overridden by Chart().namespace if set
+        - default_namespace_management_settings (dictionary). Passed in but will be overridden by Chart().namespace_management_settings if set
         """
 
         # Set the namespace
         if self.namespace is None:
-            self._namespace = namespace
+            self._namespace = default_namespace
+
+        # Set the namespace-managment settings
```
sudermanjr:
```suggestion
        # Set the namespace-management settings
```

reckoner/tests/test_course.py
```diff
@@ -229,5 +230,85 @@ def test_course_git_repository_handling(self, mockGetHelm, mockYAML):
...
+        # expect the first chart install to bubble up an error
+        chart = mock.MagicMock()
+        c.install_charts([chart])
+        chart.install.assert_called_with(
+            context=c.context,
+            default_namespace=c.namespace,
+            default_namespace_management=c.namespace_management
+        )
+
+    def test_course_namespace_and_managment_handling(self, mockGetHelm, mockYAML):
```
sudermanjr:
```suggestion
    def test_course_namespace_and_management_handling(self, mockGetHelm, mockYAML):
```

docs/README.md
```diff
@@ -41,6 +51,7 @@ The `charts` block in your course define all the charts you'd like to install an
 
 - `namespace` _(string)_  
     The namespace in which to install this chart (overrides root level `namespace` definition)
+- `namespace_management` _(object)_ **(optional)** a structure the defines default annotations and labels to the namespace that Reckoner will create. Structure matches the `default` structure at the top level `namespace_managment` block.
```
sudermanjr:
```suggestion
- `namespace_management` _(object)_ **(optional)** a structure the defines default annotations and labels to the namespace that Reckoner will create. Structure matches the `default` structure at the top level `namespace_management` block.
```

reckoner/chart.py
```diff
@@ -90,6 +90,7 @@ def __init__(self, chart, helm):
         self.args = []
 
         self._namespace = self._chart.get('namespace')
+        self._namespace_management = self._chart.get('namespace_management', {})
```
sudermanjr: Nice catch.

ejether:
I don't have a strong feeling about it. It was such simple change for the dumb version and I think that documenting it is sufficient but... I do not have a strong opinion and commenting out the line I just added along with removing the docs would make it like it never existed.

sudermanjr:
> I don't have a strong feeling about it. It was such simple change for the dumb version and I think that documenting it is sufficient but... I do not have a strong opinion and commenting out the line I just added along with removing the docs would make it like it never existed.

I think this is probably fine. @bambash how do you feel about it?

bambash:
hmm I wanted to say that it is fine as is - but after tinkering for a bit, I'm with Andy on holding off before implementing the chart level namespace management. I expected the behavior of `settings.overwrite: true` at the top level to overwrite duplicate keys at the chart level (without overwrite turned on), but that didn't happen. Overall, this feels usable but maybe we flesh out chart level namespaces a bit more before releasing it into the wild?

ejether:
Hmm. It was your expectation that the default values in 
```
namespace_management:
  default:
    metadata:{}
```
would override the value in 
```
chart:
  namespace_managment:
    metadata: {}
```
If so that is the opposite of what I would expect from the hierarchy 🤔 

So, like, clearly we are not on the same page.  :D 
I'll remove the support for chart level namespace_management in this PR and open a new one with it so that we can unblock this and have a discussion about use cases and implementation over there.

PR #179, opened 2020-02-18
Ejether/chart level namespace manager

ejether:
Adding back chart fxn removed from #177 

Let's discuss what we want this to actually do. 

As it stands, the top level `namespace_managment.default` is completely overridden by the chart's `namespace_management` block. What do we want it to actually do?


ejether:
This got reworked and added in another PR

PR #191, opened 2020-04-08
Update supported helm versions

bambash:
Updated tests and README to include supported versions of Helm.

### DIFF
--- a/end_to_end_testing/setup_helm3.sh
+++ b/end_to_end_testing/setup_helm3.sh
@@ -1,7 +1,7 @@
 source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
 
 echo "Installing Helm"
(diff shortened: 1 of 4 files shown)


end_to_end_testing/setup_helm3.sh
```diff
@@ -1,7 +1,7 @@
 source "$( cd -P "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"/setup_common.sh
 
 echo "Installing Helm"
```
ejether, line 3:
```suggestion
echo "Installing Helm 3.1"
```

ejether:
Is there something special about 2.12 that it needs it's own tests or can `helm 2` in this context mean that lates version of `helm 2`?

bambash:
I figured if we're calling out a minimum version to support we should include it in end to end testing. I'm fine removing the extra test if we don't find value in it.

ejether:
That is good enough logic for me. I'm good with it.

PR #192, opened 2020-04-08
skip tests from coverage

hopson:
I think this makes sense to do and gives a better picture of coverage changes.


hopson:
I'm guessing test methods like `setUp` were counting as coverage 🤷‍♀ 🤷‍♂

Issue #54, opened 2019-01-25
Local file charts should not try to `helm repo add`


sudermanjr:
If a path is specified as the repository, we see this warning which is confusing:
```
✔ ± reckoner plot course.yml --heading ip-masq-agent --dry-run
2019-01-25 09:36:19 root[36831]     INFO Installing ip-masq-agent
2019-01-25 09:36:20 root[36831]  WARNING Unable to install repository ../../../../../charts: Command Failed with output below:
STDOUT:
STDERR: Error: Could not find protocol handler for:

COMMAND: helm repo add ../../../../../charts None
2019-01-25 09:36:21 root[36831]     INFO [debug] Created tunnel using local port: '51041'

[debug] SERVER: "127.0.0.1:51041"
```

I believe we shouldn't be trying to do a `helm repo add` if the "repository" is a local file.

lucasreed:
I'm re-opening this because I think this needs to be functional. We have a local chart we want to install and it is the same behavior seen above.

```reckoner plot course.yml --heading fluentd-gelf
2020-04-01 09:22:48 Lukes-MacBook-Pro.local root[11895] INFO Found Helm Version 3.1.2
2020-04-01 09:22:52 Lukes-MacBook-Pro.local root[11895] INFO Installing fluentd-gelf
2020-04-01 09:22:53 Lukes-MacBook-Pro.local root[11895] WARNING Unable to install repository ../../../../../charts: Command Failed with output below:
STDOUT:
STDERR: Error: could not find protocol handler for:

COMMAND: helm repo add ../../../../../charts None
Traceback (most recent call last):
  File "reckoner/course.py", line 180, in install_charts
  File "reckoner/chart.py", line 285, in install
  File "reckoner/chart.py", line 256, in install
  File "reckoner/repository.py", line 80, in install
  File "reckoner/repository.py", line 77, in install
  File "reckoner/helm/client.py", line 173, in repo_add
  File "reckoner/helm/client.py", line 125, in execute
reckoner.helm.client.HelmClientException: Command Failed with output below:
STDOUT:
STDERR: Error: could not find protocol handler for:

COMMAND: helm repo add ../../../../../charts None

2020-04-01 09:22:53 Lukes-MacBook-Pro.local root[11895] ERROR Helm upgrade failed on fluentd-gelf
2020-04-01 09:22:53 Lukes-MacBook-Pro.local root[11895] ERROR ERROR: Chart failed to install.
[...]

PR #194, opened 2020-04-09
Test repository

hopson:
Add some simple repository tests for #54

### DIFF
--- a/reckoner/config.py
+++ b/reckoner/config.py
@@ -58,7 +58,7 @@ class Config(object):
                 self._config['home'] = helm_home
             else:
                 self._config['home'] = fallback_home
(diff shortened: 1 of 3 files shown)


hopson:
It's not clear to me if any of the `end-to-end-helmv2` errors are intentional, I'll try to look into it later.  The timeout seems like the main failure I think.

sudermanjr:
> It's not clear to me if any of the `end-to-end-helmv2` errors are intentional, I'll try to look into it later. The timeout seems like the main failure I think.

Yeah, I'm not sure what's up with that. It's the same error that I fixed recently, so idk. I will probably look at it too

sudermanjr:
Looks like that test is just flaky :-(

PR #195, opened 2020-04-14
Testing reckoner kubeconfig aws thing

ejether:
testing adding awscli to the setup.py to address issue where package may not be able to import from a virtualenv

Testing only for now


ejether:
Concluded testing

PR #207, opened 2020-07-11
Enable chart level Namespace Manager

ejether:
The machinery was already in place, I switched it on, added a little more in the docs, and updated the log level for when chart level namespace manangment is set to overwrite

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -85,6 +85,14 @@ charts:
     chart: nginx-ingress
     version: "0.25.1"
     namespace: nginx-ingress
+    namespace_management:
+      metadata:
+        annotations:
+          foo: bar
+        labels:
+          boo: far
+      settings:
+        overwrite: false
     repository: stable
     hooks:
       pre_install: echo hi
--- a/docs/README.md
+++ b/docs/README.md
@@ -166,6 +166,8 @@ minimum_versions:
 
 When you wish to manage annotations or labels for the namespaces you are installing into with Reckoner, this `namespace_management` block to define default namespace metadata and the whether or not it should overwrite the values that exist.
 
+`namespace_management` blocks can be defined at the top level or at the chart level. By default, the top level `default` metadata will be used for all namespaces and any `metadata.annotations` or `metadata.labels` set in the charts will be additive. However, if `settings.overwrite` is `True` then the `metadata` block from the chart will replace all overwrite any matching labels or annotations values.
+
 Example:
 ```yaml
 namespace_management:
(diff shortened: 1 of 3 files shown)


ejether:
Pursuant to the discussion about Astro yesterday, made this change.

rbren:
✨

reckoner/kube.py
```diff
@@ -91,7 +91,7 @@ def patch_metadata(self):
         """
         if self.overwrite:
             patch_metadata = self.metadata
-            logging.debug("Overwriting Namespace Metadata")
+            logging.info("Overwriting Namespace Metadata")
```
sudermanjr, line 94: Will this include the namespace that is having its metadata overwritten? If not, then I think we should add that info to the message

sudermanjr: Also, perhaps a debug or info message when we refuse to overwrite because it is false. That way the user isn't surprised when nothing happens. Is default for overwrite `false` ?

ejether: Truncated output from the last changes. 

```
# reckoner plot course.yml -o a1 -o a2
2020-07-13 14:33:09 MacBook-Pro.local root[49384] INFO Installing a1
2020-07-13 14:33:09 MacBook-Pro.local root[49384] INFO Overwriting Namespace 'foobear' Metadata
2020-07-13 14:33:11 MacBook-Pro.local root[49384] INFO Installing a2
2020-07-13 14:33:11 MacBook-Pro.local root[49384] INFO Not Overwriting Metadata Label 'ManagedBy' in Namespace 'foobear'
```

docs/README.md
```diff
@@ -158,6 +166,8 @@ minimum_versions:
 
 When you wish to manage annotations or labels for the namespaces you are installing into with Reckoner, this `namespace_management` block to define default namespace metadata and the whether or not it should overwrite the values that exist.
 
+`namespace_management` blocks can be defined at the top level or at the chart level. By default, the top level `default` metadata will be used for all namespaces and any `metadata.annotations` or `metadata.labels` set in the charts will be additive. However, if `settings.overwrite` is `True` then the `metadata` block from the chart will replace all overwrite any matching labels or annotations values.
```
sudermanjr, line 169:
```suggestion
`namespace_management` blocks can be defined at the top level or at the chart level. By default, the top level `default` metadata will be used for all namespaces and any `metadata.annotations` or `metadata.labels` set in the charts will be additive. However, if `settings.overwrite` is `True` then the `metadata` block from the chart will replace any matching labels or annotations values.
```

sudermanjr: Also, I think we should document the fact that the chart-level namespace_management block cannot `remove` or `delete` course-level annotations/labels. It can only `modify` them.

ejether:
@bambash I missed @sudermanjr 's comment requesting your review on this before merging. I'm happy to make any fixes/changes you'd like here so please still give it a look and let me know and I'll revert or create a new PR to address those changes.

Issue #186, opened 2020-03-26
Introduce --all flag and make default behavior run a single heading


lucasreed:
Title is fairly self-explanatory, but this is a major change since folks outside of Fairwinds may be using this, and it could cause some disruption. 

Therefore this will be a major version release change methinks.

sudermanjr:
I think it would ease the change if we made it something like this:

```
reckoner plot course.yml
error: --heading or --all are required
```

mgoodhart5:
I think that keeping `-o` as an option (or something that is short like that) instead of having to type `--heading` every time would be nice. It is way faster to type!

PR #208, opened 2020-07-13
Add Helm Template

ejether:
This PR adds the `template` command through Reckoner. Since a lot of the machinery needed for `plot` was reusable, there was a fair amount of refactoring to keep things dry-er. Template itself just `print`s the tempalte output to stdout instead of logging because logging in python goes to STDERR. This means you can `reckoner template -o <chart> course.yml > manifests.yml` and you'll get a file with yaml and still see all the logs. 

I elected to make `-o` required thinking that one probably does not want to template out the entire reckoner file but I'm on the fence about it. Happy to hear feedback.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -182,7 +182,9 @@ Options:
   --help            Show this message and exit.
 
 Commands:
-  plot     Install charts with given arguments as listed in yaml file...
+  plot     Install charts with given arguments as listed in yaml file argument
+  template Output the template of the chart or charts as they would be installed or
+  upgraded
   version  Takes no arguments, outputs version info
 ```
 
@@ -211,4 +213,26 @@ Options:
                                   does not already exist. Replaces
                                   functionality lost in Helm3
   --help                          Show this message and exit.
+```
+
+Or
+```text
+# reckoner template --help
+Usage: reckoner template [OPTIONS] COURSE_FILE
+
+  Output the template of the chart or charts as they would be installed or
+  upgraded
+
+Options:
+  -o, --only, --heading <chart>  Only run a specific chart by name  [required]
+  --helm-args TEXT               Passes the following arg on to helm, can be
+                                 used more than once. WARNING: Setting this
+                                 will completely override any helm_args in the
+                                 course. Also cannot be used for configuring
+                                 how helm connects to tiller.
+
+  --log-level TEXT               Log Level. [INFO | DEBUG | WARN | ERROR].
+                                 (default=INFO)
+
+  --help                         Show this message and exit.
 ```
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -220,16 +220,7 @@ class Chart(object):
             nsm = NamespaceManager(self.namespace, self.namespace_management)
             nsm.create_and_manage()
 
-    def install(self, default_namespace=None, default_namespace_management={}, context=None) -> None:
-        """
-        Description:
-        - Upgrade --install the course chart
-
-        Arguments:
-        - default_namespace (string). Passed in but will be overridden by Chart().namespace if set
-        - default_namespace_management_settings (dictionary). Passed in but will be overridden by Chart().namespace_management_settings if set
-        """
-
+    def __pre_command(self, default_namespace=None, default_namespace_management={}, context=None) -> None:
         # Set the namespace
         if self.namespace is None:
             self._namespace = default_namespace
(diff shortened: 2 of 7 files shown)


sudermanjr:
RE: making -o required 
https://github.com/FairwindsOps/reckoner/issues/186

docs/README.md
```diff
@@ -182,7 +182,9 @@ Options:
   --help            Show this message and exit.
 
 Commands:
-  plot     Install charts with given arguments as listed in yaml file...
+  plot     Install charts with given arguments as listed in yaml file argument
```
sudermanjr, line 185: These end up coming across as:

```
  plot      Install charts with given arguments as listed in yaml file...
  template  Output the template of the chart or charts as they would be...
  version   Takes no arguments, outputs version info
```

Do we need to make them shorter?

reckoner/chart.py
```diff
@@ -243,28 +234,39 @@ def install(self, default_namespace=None, default_namespace_management={}, conte
             self._context = context
         # Try to run the install process for mark the result as failed
 
-        try:
-            self.manage_namespace()
+        # TODO: Improve error handling of a repository installation
```
sudermanjr, line 237: Should we open an issue for this in order to track it?

reckoner/chart.py
```diff
@@ -243,28 +234,39 @@ def install(self, default_namespace=None, default_namespace_management={}, conte
...
-            #       way we could find out earlier our course is wrong.
-            self.repository.install(self.name, self.version)
+        # Build the args for the chart installation
+        # And add any extra arguments
+        self.build_helm_arguments_for_chart()
 
-            # Update the helm dependencies
-            self.update_dependencies()
+        # Check and Error if we're missing required env vars
+        # TODO Rename this function as it does more than just "check", it also interpolates
```
sudermanjr, line 251: I think we can drop this TODO since we're probably not gonna do that (I know you were just carrying it over)

reckoner/cli.py
```diff
@@ -53,8 +53,10 @@ def cli(ctx, log_level, *args, **kwargs):
               help="Attempt to install all charts in the course, even if any charts or hooks fail to run.")
 @click.option("--create-namespace/--no-create-namespace", default=True,
               help="Will create the specified nameaspace if it does not already exist. Replaces functionality lost in Helm3")
-def plot(ctx, course_file=None, dry_run=False, debug=False, only=None, helm_args=None, continue_on_error=False, create_namespace=True):
+@click.option("--log-level", default="INFO", help="Log Level. [INFO | DEBUG | WARN | ERROR]. (default=INFO)")
```
sudermanjr, line 56: 🎉

reckoner/course.py
```diff
@@ -199,37 +198,83 @@ def install_charts(self, charts_to_install: list) -> List[ChartResult]:
...
+                'None of the charts you requested ({}) could be '
+                'found in the course list. Verify you are using the '
+                'release-name and not the chart name.'.format(', '.join(charts_requested))
+            )
+
+        return self._only_charts
+
+    def template(self, charts_requested_to_template: list) -> List[str]:
+        """
+        Accepts charts_to_install, an interable of the names of the charts
```
sudermanjr, line 259:
```suggestion
        Accepts charts_requested_to_template, an iterable of the names of the charts
```

reckoner/course.py
```diff
@@ -199,37 +198,83 @@ def install_charts(self, charts_to_install: list) -> List[ChartResult]:
...
+        to install. This method compares the charts in the argument to the
```
sudermanjr, line 260:
```suggestion
        to template. This method compares the charts in the argument to the
```

reckoner/course.py
```diff
@@ -199,37 +198,83 @@ def install_charts(self, charts_to_install: list) -> List[ChartResult]:
...
+        charts in the course and calls Chart.template()
+
+        """
+        # return the text of the charts templating
+        results = self.template_charts(self.only_charts(charts_requested_to_template))
+        return results
 
+    def plot(self, charts_requested_to_install: list) -> List[ChartResult]:
+        """
+        Accepts charts_to_install, an interable of the names of the charts
```
sudermanjr, line 270:
```suggestion
        Accepts charts_to_install, an iterable of the names of the charts
```

reckoner/reckoner.py
```diff
@@ -106,6 +106,19 @@ def install(self, charts: List[str] = []) -> None:
             logging.error(error)
             raise ReckonerCommandException('Failed to find any valid charts to install.')
 
+    def template(self, charts: List[str] = [] ):
+        selected_charts = charts or [chart._release_name for chart in self.course.charts]
+        try:
+            return self.course.template(selected_charts)
+            # for chart_result in plot_results:
```
sudermanjr, line 113: I think these commented lines can be removed

docs/README.md
ejether: I don't think so because if you do `reckoner plot --help` you get the full string 
You only get the truncated string with you do `reckoner --help`

I chose to copy/pasta the long string into the docs to make it easier to read them though.

reckoner/chart.py
ejether: https://github.com/FairwindsOps/reckoner/issues/209

docs/README.md
sudermanjr: Ah, cool. Seems fine then

bambash:
awesome! working great locally testing. I like the `-o` requirement, but could we add in a way to pass a `-a` or `--all` flag?

ejether:
@bambash We're going to add the `--all` in another PR addressing #186 specifically.
Do you think that's ok or do you think it really should be here?

bambash:
:+1: I'm fine if we wait to add an `--all` flag before creating the next release

PR #210, opened 2020-07-14
MAJOR (potentially breaking)  Add --run-all. Make either run-all or only required.

### LINKED ISSUE #186: Introduce --all flag and make default behavior run a single heading

sudermanjr:
Makes either --run-all or --only required. Running `reckoner plot course.yml` with no arguments will now produce an error.

This applies to both the `plot` and `template` commands

```
  -a, --run-all                   Run all charts in the course. Mutually
                                  exclusive with 'only'.

  -o, --only, --heading <chart>   Only run a specific chart by name Mutually
                                  exclusive with 'run_all'.
```

```
└─ reckoner plot course.yaml -o foo -a
Error: Illegal usage: 'only' is mutually exclusive with 'run_all'.
```

Fixes #186

### DIFF
--- a/reckoner/cli.py
+++ b/reckoner/cli.py
@@ -26,6 +26,24 @@ from .reckoner import Reckoner
 
 from reckoner.schema_validator.course import validate_course_file
 
+class Mutex(click.Option):
+    def __init__(self, *args, **kwargs):
+        self.not_required_if:list = kwargs.pop("not_required_if")
+
+        assert self.not_required_if, "'not_required_if' parameter required"
+        kwargs["help"] = (kwargs.get("help", "") + " Mutually exclusive with '" + ", ".join(self.not_required_if) + "'.").strip()
+        super(Mutex, self).__init__(*args, **kwargs)
+
+    def handle_parse_result(self, ctx, opts, args):
+        current_opt:bool = self.name in opts
+        for mutex_opt in self.not_required_if:
+            if mutex_opt in opts:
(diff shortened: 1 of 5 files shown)


reckoner/cli.py
```diff
@@ -114,4 +146,4 @@ def template(ctx, only, log_level, course_file=None, helm_args=None,):
 @cli.command()
 def version():
     """ Takes no arguments, outputs version info"""
-    print(__version__)
+    print(__version__)
```
ejether, line 149:
```suggestion
    print(__version__)

```

reckoner/cli.py
```diff
@@ -90,20 +114,28 @@ def plot(ctx, log_level, course_file=None, dry_run=False, debug=False, only=None
...
+            logging.error("You must pass either --run-all or --only.")
+            ctx.exit(1)
+    
     # Check Schema of Course FileA
     with open(course_file.name, 'rb') as course_file_stream:
         validate_course_file(course_file_stream)
     # Load Reckoner
     r = Reckoner(course_file=course_file, helm_args=helm_args)
-    # Convert tuple to list
+    # Convert tuple to list
```
ejether, line 138:
```suggestion
    # Convert tuple to list
```

PR #212, opened 2020-07-15
Adding `get-manifests` command

ejether:


ejether:
This is branched from https://github.com/FairwindsOps/reckoner/pull/210

So should not be merged until that is.

reckoner/helm/provider.py
```diff
@@ -42,7 +42,12 @@ def execute(cls, helm_command):
 
         # start by creating a command line arguments list with the command being first
         args = list([instance._helm_binary])
-        args.append(instance._helm_command.command)
+
+        # if command has a space in it (like get manifests), split on space
+        # and append each segment as it's own list item to make `call` happy
+        for command_segment in instance._helm_command.command.split(' '):
+            args.append(command_segment)
+
```
ejether, line 50: This is a potentially risky change.

ejether:
Ready for review.

reckoner/chart.py
```diff
@@ -295,6 +295,34 @@ def template(self, default_namespace=None, default_namespace_management={}, cont
         finally:
             self.clean_up_temp_files()
 
+    def get_manifest(self, default_namespace=None, default_namespace_management={}, context=None) -> None:
+        self.__pre_command(default_namespace, default_namespace_management, context)
+        try:
+            # get_manifest needs a differnt set of args
```
sudermanjr, line 301:
```suggestion
            # get_manifest needs a different set of args
```

ejether:
I think, the only current scenario where that provider change would bite us is if, when we implement another helm command, the person doing the code adds extra spaces between the command terms. 
`get    manifests`  would end up like `['get',' ',' ','manifests']` in the args.

Seems unlikely

PR #213, opened 2020-07-17
Python3.8 testing

ejether:
adding python 3.8 test build


ejether:
bigger can than I wanted.

PR #214, opened 2020-07-22
Fixing template and get-manifests

ejether:
I realized that there were some mistakes in the get-manifest PR and this should address them

### DIFF
--- a/reckoner/course.py
+++ b/reckoner/course.py
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
+import traceback
 import logging
 import semver
 import sys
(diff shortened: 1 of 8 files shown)


reckoner/tests/test_cli.py
```diff
@@ -237,7 +237,7 @@ def test_template_handles_exception(self, reckoner_mock):
 
         self.assertEqual(1, result.exit_code, result.output)
 
-    def test_template_options(self):
+    def test_get_manefest_options(self):
```
sudermanjr, line 240:
```suggestion
    def test_get_manifest_options(self):
```

reckoner/reckoner.py
```diff
@@ -109,6 +109,7 @@ def install(self, charts: List[str] = []) -> None:
     def template(self, charts: List[str] = [] ):
         selected_charts = charts or [chart._release_name for chart in self.course.charts]
         try:
+            print(selected_charts)
```
sudermanjr, line 112: possible debug statement?

ejether:
```suggestion
```

ejether, line 112:
```suggestion
```

reckoner/course.py
```diff
@@ -228,13 +225,10 @@ def get_chart_manifests(self, charts_to_manifest: list) -> List[str]:
...
                     context=self.context
                 )
+                results.append(command_response)
             except (Exception, ReckonerCommandException) as e:
                 import traceback
-                logging.debug(print(traceback.format_exc()))
-                raise e
-            finally:
-                # Always grab any results in the chart results
-                results.append(command_response)
```
sudermanjr, line 237: Why aren't we keeping track of the results for template and manifest like we do for plot?

ejether: `plot` is setup to be able to continue when one chart fails.  
The `chart` instance exists in the scope of the `for chart in` loop so when `finally` appends `chart.result` to `results` it exists.

Since `.template` and `.get_manifests` return a string (currently) instead of a `ChartResult` instance the `command_response` variable only exists in the scope of the `try` block and if it throws an error, the run stops. 

I took the lazy path and moved the `results.append` into the `try` block so the `command_response` would be in scope. 

Looking at it again, I could line it up more with the `install` method and use the `ChartResult` class by adding an `output` then printing the output in the `cli`. 

I assumed no one would want a partial successful run of `template` or `get-manifests` but I can easily implement as above if you think it is a better pattern.

sudermanjr: I _think_ that a partial response from template, along with capturing errors will still be useful.  I imagine someone using this to validate that their course.yml is behaving as expected. 

My gut says that keeping things aligned will be better if it's not too much additional work, but I'm not overly concerned about it.

ejether: Yeah, I think you're right and it won't be too much work.

PR #217, opened 2020-08-11
Adding Diff command

ejether:
This adds `diff` to the list of commands. It compares
   the output of `template`  and  `get_manifests`  using `difflib` . Whereas
   most other commands return a `HelmCommandResponse` response, that was not
   really work able without major re-write here because there is only a single
   instance of `self.response`  in the  class. Calling `template`  and
   `get_manifests` in succession from an instnace of the Chart class
   overwrites the previous command's `self.response` . Less than ideal, but I
   created hidden methods to return the content of those commands that are
   then called by the new `diff`  command and I return the string
   representation of the diff as `self.resposne`  but it is not an instance of
   HelmResponse becaue it is not, technically, a HelmResponse

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -199,7 +199,7 @@ Commands:
   upgraded
   get-manifests  Output the manifests of the chart or charts as they are installed
   diff           Output diff of the templates that would be installed and the manifests as they are installed
-  update         Checks to see if any thing will have changed, if so, updated the release, if not, does nothing
+  update         Checks to see if anything will be changed, if so, update the release, if not, does nothing
   version  Takes no arguments, outputs version info
 ```
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -194,9 +194,12 @@ Options:
   --help            Show this message and exit.
 
 Commands:
-  plot     Install charts with given arguments as listed in yaml file argument
-  template Output the template of the chart or charts as they would be installed or
+  plot           Install charts with given arguments as listed in yaml file argument
+  template       Output the template of the chart or charts as they would be installed or
   upgraded
+  get-manifests  Output the manifests of the chart or charts as they are installed
+  diff           Output diff of the templates that would be installed and the manifests as they are installed
+  update         Checks to see if any thing will have changed, if so, updated the release, if not, does nothing
   version  Takes no arguments, outputs version info
 ```
 
@@ -206,6 +209,49 @@ $> reckoner plot --help
 Usage: reckoner plot [OPTIONS] COURSE_FILE
 
   Install charts with given arguments as listed in yaml file argument
+
+Options:
+  --dry-run                       Pass --dry-run to helm so no action is
+                                  taken. Implies --debug and skips hooks.
+
+  --debug                         DEPRECATED - use --log-level=DEBUG as a
+                                  parameter to `reckoner` instead. May be used
+                                  with or without `--dry-run`. Or, pass
+                                  `--debug` to --helm-args
+
+  -a, --run-all                   Run all charts in the course. Mutually
+                                  exclusive with 'only'.
+
+  -o, --only, --heading <chart>   Only run a specific chart by name Mutually
+                                  exclusive with 'run_all'.
+
+  --helm-args TEXT                Passes the following arg on to helm, can be
+                                  used more than once. WARNING: Setting this
+                                  will completely override any helm_args in
+                                  the course. Also cannot be used for
+                                  configuring how helm connects to tiller.
+
+  --continue-on-error             Attempt to install all charts in the course,
+                                  even if any charts or hooks fail to run.
+
+  --create-namespace / --no-create-namespace
+                                  Will create the specified nameaspace if it
+                                  does not already exist. Replaces
+                                  functionality lost in Helm3
+
+  --log-level TEXT                Log Level. [INFO | DEBUG | WARN | ERROR].
+                                  (default=INFO)
+
+  --help                          Show this message and exit.
+```
+
+Or
+```
+# reckoner update --help
+Usage: reckoner update [OPTIONS] COURSE_FILE
+
+  Checks to see if any thing will have changed, if so, updated the release,
+  if not, does nothing
 
 Options:
   --dry-run                       Pass --dry-run to helm so no action is
--- a/docs/README.md
+++ b/docs/README.md
@@ -250,7 +250,7 @@ Or
 # reckoner update --help
 Usage: reckoner update [OPTIONS] COURSE_FILE
 
-  Checks to see if any thing will have changed, if so, updated the release,
+  Checks to see if anything will be changed, if so, update the release,
   if not, does nothing
 
 Options:
--- a/end_to_end_testing/run_end_to_end.sh
+++ b/end_to_end_testing/run_end_to_end.sh
@@ -573,6 +573,33 @@ function e2e_test_get_manifests_non_existent() {
     fi
 }
 
+function e2e_test_get_diff() {
+    if [ "${HELM_VERSION}" -eq "3" ]; then
+        reckoner plot test_basic.yml --run-all
+        if ! reckoner get-manifests test_basic.yml -o first-chart; then
+            mark_failed "${FUNCNAME[0]}" "Expected to succeed getting diff of release"
+        fi
+    fi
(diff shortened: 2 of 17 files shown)


sudermanjr:
Went to test this out and ran into an interesting case.  If there are tests in the chart, they get shown as an addition:

```
*** 

--- 

***************

*** 159,165 ****

                memory: 500Mi
              requests:
                cpu: 100m
!               memory: 300Mi
    volumeClaimTemplates:
    - metadata:
        name: atlantis-data
--- 159,165 ----

                memory: 500Mi
              requests:
                cpu: 100m
!               memory: 500Mi
    volumeClaimTemplates:
    - metadata:
        name: atlantis-data
***************

*** 198,200 ****

--- 198,239 ----

              backend:
                serviceName: atlantis
                servicePort: 80
+ ---
+ # Source: atlantis/templates/tests/test-atlantis-pod.yaml
+ apiVersion: v1
+ kind: Pod
+ metadata:
+   name: "atlantis-ui-test-p6kyj"
+   annotations:
+     "helm.sh/hook": test-success
+ spec:
+   initContainers:
+     - name: test-framework
+       image: dduportal/bats:0.4.0
+       command:
+       - "bash"
+       - "-c"
+       - |
+         set -ex
+         # copy bats to tools dir
+         cp -R /usr/local/libexec/ /tools/bats/
+       volumeMounts:
+       - mountPath: /tools
+         name: tools
+   containers:
+     - name: atlantis-ui-test
+       image: lachlanevenson/k8s-kubectl:v1.4.8-bash
+       command: ["/tools/bats/bats", "-t", "/tests/run.sh"]
+       volumeMounts:
+       - mountPath: /tests
+         name: tests
+         readOnly: true
+       - mountPath: /tools
+         name: tools
+   volumes:
+   - name: tests
+     configMap:
+       name: atlantis-tests
+   - name: tools
+     emptyDir: {}
+   restartPolicy: Never
```

not sure that's an issue or a blocker, but could be annoying.

Also, is there any way to tell which file the changes would be in?  I see my change, but I'm not certain where that is happening.

ejether:
Interesting with the tests... 

The `difflib` is a PIA. I can output the WHOLE thing easily with changes in line, or just the lines that are different without much effort.  Since these aren't "file" perse, they are text streams, it's not currently discrete in that way.

sudermanjr:
> Interesting with the tests...
> 
> The `difflib` is a PIA. I can output the WHOLE thing easily with changes in line, or just the lines that are different without much effort. Since these aren't "file" perse, they are text streams, it's not currently discrete in that way.

I had a feeling that might be the case.  I'm good with this as a first implementation.  Perhaps in the future we can yaml parse it into documents and then diff those individually.

docs/README.md
```diff
@@ -194,9 +194,12 @@ Options:
...
 
 Commands:
-  plot     Install charts with given arguments as listed in yaml file argument
-  template Output the template of the chart or charts as they would be installed or
+  plot           Install charts with given arguments as listed in yaml file argument
+  template       Output the template of the chart or charts as they would be installed or
   upgraded
+  get-manifests  Output the manifests of the chart or charts as they are installed
+  diff           Output diff of the templates that would be installed and the manifests as they are installed
+  update         Checks to see if any thing will have changed, if so, updated the release, if not, does nothing
```
sudermanjr, line 202:
```suggestion
  update         Checks to see if anything will have changed, if so, update the release, if not, does nothing
```

docs/README.md
```diff
@@ -242,6 +245,49 @@ Options:
   --help                          Show this message and exit.
 ```
 
+Or
+```
+# reckoner update --help
+Usage: reckoner update [OPTIONS] COURSE_FILE
+
+  Checks to see if any thing will have changed, if so, updated the release,
```
sudermanjr, line 253:
```suggestion
  Checks to see if anything will be changed, if so, update the release,
```

ejether:
A chart with tests will always be `plot`ed if you youse the `update` command 🤔 Probably should sort that out.

sudermanjr:
> A chart with tests will always be `plot`ed if you youse the `update` command 🤔 Probably should sort that out.

I was just about to say that. I hadn't actually tried the `update` command yet

sudermanjr:
Couple more things:

1) Helm hooks that have a lifecycle policy will be affected the same way as tests:

(Truncated)
```
***************

*** 173,175 ****

--- 173,248 ----

              backend:
                serviceName: goldilocks-dashboard
                servicePort: http
+ ---
+ # Source: goldilocks/templates/vpa-install-hook.yaml
+ apiVersion: v1
+ kind: ServiceAccount
+ metadata:
+   annotations:
+     "helm.sh/hook": "pre-install,pre-upgrade"
+     "helm.sh/hook-delete-policy": "hook-succeeded,before-hook-creation,hook-failed"
+     "helm.sh/hook-weight": "-100"
+   name: goldilocks-vpa-install
```

2) When running `reckoner diff course.yml -a`, I get all the diffs at the end with no indication which chart they came from.  The INFO output says that it ran the diff, but does not indicate that there were changes. Then the diffs are all dumped one-by-one.  Maybe just adding a line at the top of the diff output that says `diff for chart/release blah` would be sufficient, and adding an INFO log message that says `A diff was found`

(partial outptu)
```
2020-08-14 11:34:18  root[75853]     INFO  Running 'diff' on goldilocks
2020-08-14 11:34:19  root[75853]     INFO  Running 'diff' on atlantis





***

---

***************

*** 231,233 ****

--- 231,255 ----

    insecureSkipTLSVerify: true
    groupPriorityMinimum: 100
    versionPriority: 100
+ ---
+ # Source: metrics-server/templates/tests/test-version.yaml
+ apiVersion: v1
+ kind: Pod
+ metadata:
+   name: metrics-server-test
+   labels:
```

ejether:
@sudermanjr I think I've addressed all the diff issues. Tests need writing and fixing but if you have time to give the commands a run through while I write the tests, I'd appreciate it.

sudermanjr:
I just installed from your branch and tried it out on a chart that shouldn't have any changes:

```
2020-08-25 10:13:30  root[31168]    DEBUG  helm template cert-manager jetstack/cert-manager --namespace cert-manager --version v0.14.3 -f /var/folders/hv/vcbnq0_d11n8ss9dsxcx66w00000gn/T/tmp47w7pxof.yml
2020-08-25 10:13:32  root[31168]    DEBUG  /var/folders/hv/vcbnq0_d11n8ss9dsxcx66w00000gn/T/tmpzpvtcs7q.yml not found to delete. Ignoring this fact
2020-08-25 10:13:32  root[31168]    DEBUG  manifest ServiceAccount named cert-manager-cainjector has no annotations to filter on
2020-08-25 10:13:32  root[31168]    DEBUG  Traceback (most recent call last):
  File "/Users/asuderma/.cthulhu/cuddle/lib/python3.7/site-packages/reckoner/course.py", line 180, in __run_command_for_charts_list
    context=self.context
  File "/Users/asuderma/.cthulhu/cuddle/lib/python3.7/site-packages/reckoner/chart.py", line 411, in diff
    self.result.response = self.__diff_response()
  File "/Users/asuderma/.cthulhu/cuddle/lib/python3.7/site-packages/reckoner/chart.py", line 393, in __diff_response
    diff = manifestDiff(manifest_response, template_response)
  File "/Users/asuderma/.cthulhu/cuddle/lib/python3.7/site-packages/reckoner/manifests.py", line 161, in diff
    _manifest.filter_by_annotation_name('helm.sh/hook')
  File "/Users/asuderma/.cthulhu/cuddle/lib/python3.7/site-packages/reckoner/manifests.py", line 113, in filter_by_annotation_name
    if filter_annotation_key not in [manifest_annotation_key for manifest_annotation_key in manifest.annotations]:
TypeError: 'NoneType' object is not iterable

2020-08-25 10:13:32  root[31168]    ERROR  ERROR: diff Failed on cert-manager
2020-08-25 10:13:32  root[31168]    ERROR  Stopping 'diff' for chart due to an error! Some of your requested actions may not have been completed!
2020-08-25 10:13:32  root[31168]    ERROR  'NoneType' object is not iterable
None
```

ejether:
merging #219 prevented that one last warning we ran into about `helm init`
I believe this is ready to go now.

PR #220, opened 2020-09-03
Fix binary compilation. Add a python 3.8 build test

sudermanjr:
Darwin binary - https://3115-159374996-gh.circle-artifacts.com/0/reckoner-darwin
Linux Binary - https://3114-159374996-gh.circle-artifacts.com/0/reckoner-linux-amd64


sudermanjr:
@microcosem  or @bambash if you could test this on Linux I would really appreciate that.

ejether (approved):

PR #222, opened 2020-09-04
Adjust dry-run behavior for update, get-manifest, template

### LINKED ISSUE #221: `update --dry-run` always shows that and update is needed
when `--dry-run` is set for `update`  the call to `template` also gets `--dry-run` set and it outputs the full template whereas `get manifest` outputs nothing so it will always see a difference. When `--dry-run` is not set, `update` proceeds as expected.

ejether:
removing dryrun from reckoner tempalte, its not necessary, removed dryrun from update, also unnecsasary, still gated the debug args in the get-manifest flow because it breaks stuff if its there

Fixed #221

### DIFF
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -306,6 +306,7 @@ class Chart(object):
 
         try:
             if self.requires_update:
+                logging.info(f"Release {self.name} requires updating.")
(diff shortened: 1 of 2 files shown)


sudermanjr (approved):
LGTM

PR #224, opened 2020-09-15
Move e2e tests to use Venom. Start deprecating Helm2

sudermanjr:
Refactored all e2e tests to use https://github.com/ovh/venom. Also changed the setup to use the rok8s orb and a kind cluster rather than the old manual setup

I also removed all helm2 end to end testing. Helm2 will be deprecated within the year, and we need to move towards deprecation as well. https://helm.sh/blog/helm-v2-deprecation-timeline/

TODO: 
- [x] Duplicate the rest of the tests in the old run_end_to_end.sh script
- [x] Merge and release https://github.com/FairwindsOps/rok8s-scripts/pull/366 so that we can use the pre-built binary rather than having to install python3.8 from source. Should speed up the tests considerably


ejether (approved):
it's a lot, but seems good.

PR #237, opened 2020-10-22
docs/README.md : Add a minor comment on git sparse checkout

elthariel:
I'm using a helm chart with a path different from its name and ended up a bit confused. After reading the source code a tiny bit, I figured it out and suggest adding a tiny comment in the doc about it


elthariel:
I realize now there was already a related comment in the chart section. Feel free to close if you think it's enough

ejether (approved):
That's a great comment @elthariel! We've had a hard time adequately describing how it works. This is very helpful.

PR #245, opened 2020-11-04
Respect Default Course Namespace for `diff` Command

ejether:
This PR ensurse that `diff` will respect the course level namespace. `args` were not being passed through properly. this PR fixes that for `diff` and `update` and addes e2e tests for `diff`


sudermanjr (approved):
LGTM. Thanks!

PR #247, opened 2020-11-06
Replace Deprecated URLs in Tests

ejether:
This PR replaces all deprecated chart urls with approved versions in the test course.yml


sudermanjr (approved):
Thanks

Issue #248, opened 2020-11-10
Make repository path relative to yaml file


elthariel:
Imagine a folder structure like this one:

```
charts/
  my-chart/
    [...]
  dev/
    plot.yml
  prod:
    plot.yml
```

If in dev/plot.yml, you have

```yaml
charts:
  my-chart:
    chart: my-chart
    repository: ../charts
```

and you run reckoner from the top-level folder, it won't find the chart folders. i.e. when using a path to specify the repository, this path is not relative to the reckoner file. I think it should. WDYT ?

elthariel:
Would you accept a PR on this ?

sudermanjr:
I believe you can just specify `chart: ../charts/dev` to make this work, right?

elthariel:
@sudermanjr It doesn not

elthariel:
If I were to place the path to the chart in the `chart` key instead of the chart name, what should I place in the `repository` field ? If i leave it empty, the default repo is applied, which isn't what is wanted. If I set an empty string there, reckoner (rightfully) fails.

The solution is to check if the repository field is a path and make it relative to the reckoner file, like the `files` key

sudermanjr:
You are 100% correct, I was mistaken on that. (been a long time since I used local charts).

Seems like that would be the correct solution, to check for a file path and make it relative to the course file like the files key. We would definitely love a PR on that if it's something you wanted to work on.

ejether:
@elthariel Good catch. The path is relative to the current working directly of the `reckoner plot` command. 
I think making it relative to the course file, as you suggest, makes a lot more sense. 

As a current work around, if you `cd charts/dev && reckoner plot dev.yml --run-all` it should work as expected. 

Let me know if you're going to submit a PR, otherwise I should have some time to make a patch today or early next week.

elthariel:
Thanks @sudermanjr, I got pulled into another project and didn't find the time to fix this

PR #251, opened 2020-11-16
Fix 248 chart relative to course file

ejether:
This PR Fixes https://github.com/FairwindsOps/reckoner/issues/248

This makes the chart repository path for local charts so that it is relative to the course file. The hooks were already handled in another way.


sudermanjr (approved):
Awesome possum. Thanks!

ejether:
Hmm, looks like hooks execution might be bugged with this fix. Hold off.

sudermanjr:
> Hmm, looks like hooks execution might be bugged with this fix. Hold off.

I would never merge before tests pass :-D

ejether:
Done

PR #265, opened 2020-12-23
[BREAKING] Deprecate binary build and installation method

sudermanjr:


end_to_end_testing/run.sh
```diff
@@ -5,8 +5,12 @@ set -e
 
 python --version
 
-chmod +x /usr/local/bin/reckoner
-reckoner version
+apt-get update
+apt-get install python3-pip
+
+cd /reckoner
+pip3 install reckoner
```
lucasreed, line 12: This is installing the most recent version from pypi, I think we want to do a `pip3 install .` here, but I could be wrong.

lucasreed (changes requested):
One in-line comment, otherwise looks good.

end_to_end_testing/run.sh
sudermanjr: Good catch.
```suggestion
pip3 install .
```

lucasreed (approved):

ejether (approved):

PR #275, opened 2021-01-24
Fix 269 - Show correct namespaces in log

### LINKED ISSUE #269: Log message always puts the default namespace
When running reckoner I see a log message that comes [from here](https://github.com/FairwindsOps/reckoner/blob/master/reckoner/course.py#L175) but the namespace that is logged out is always the default toplevel namespace even if a release is deployed somewhere else.

For instance I get this output:
```
❯ reckoner diff -o rbac-manager course.yml
2021-01-13 11:41:13 niflheim.local root[99923] INFO Found Helm Version 3.3.4
2021-01-13 11:41:22 niflheim.local root[99923] INFO Running 'diff' on rbac-manager in infra
2021-01-13 11:41:25 niflheim.local root[99923] INFO There are no differences in release rbac-manager
```

Even though the release is defined like so in `course.yml`:

```
  charts:
    rbac-manager:
      version: 1.7.0
      namespace: rbac-manager
      repository: fairwinds-stable
      values:
        priorityClassName: fairwinds-critical
```

ejether:
Gets namespace from the chart and defaults to the course level one for log display

Fixes #269

### DIFF
--- a/reckoner/course.py
+++ b/reckoner/course.py
@@ -172,7 +172,8 @@ class Course(object):
     def __run_command_for_charts_list(self, command: str, charts: list) -> List[ChartResult]:
         results = []
         for chart in charts:
-            logging.info(f"Running '{command}' on {chart.release_name} in {self.namespace}")
+            namespace = chart.namespace or self.namespace
+            logging.info(f"Running '{command}' on {chart.release_name} in {namespace}")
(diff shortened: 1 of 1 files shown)


sudermanjr (approved):
Awesome, thanks!

PR #302, opened 2021-02-15
Validate incoming args

### LINKED ISSUE #295: Validate Incoming Args
--run-all and --only should be mutually exclusive and validated before running any <plot|diff|get|template>

Write a validation function in the `cmd/root.go` file to ensure this.

Should be of the form:

```
func validateArgs(cmd *cobra.Command, args []string) error {
        // DO VALIDATION
	return nil
}
```

Note that `validateCourseFileArg` already exists and may also need to be called from that function (maybe?)

shaswa:
- make -run-all and --only mutually exclusive 
- Formatting changes my IDE complained about 

Fixes #295

### DIFF
--- a/pkg/reckoner/validation.go
+++ b/pkg/reckoner/validation.go
@@ -6,7 +6,7 @@ import (
 )
 
 func ValidateArgs(runAll bool, onlyRun, args []string) (string, error) {
-	if runAll == true && len(onlyRun) != 0 {
+	if runAll && len(onlyRun) != 0 {
 		return "", fmt.Errorf("you must either use run-all or only")
 	}
 
--- a/pkg/reckoner/validation.go
+++ b/pkg/reckoner/validation.go
@@ -1,0 +1,22 @@
+package reckoner
+
+import (
+	"fmt"
+	"os"
+)
+
+func ValidateArgs(runAll bool, onlyRun, args []string) (string, error) {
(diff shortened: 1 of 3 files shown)


pkg/reckoner/validation_test.go
```diff
@@ -0,0 +1,65 @@
...
+package reckoner
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateArgs(t *testing.T) {
```
sudermanjr, line 24:
```suggestion
// ValidateArgs validates the reckoner command arguments
func TestValidateArgs(t *testing.T) {
```

sudermanjr (changes requested):

pkg/reckoner/validation.go
```diff
@@ -0,0 +1,22 @@
+package reckoner
+
+import (
+	"fmt"
+	"os"
+)
+
+func ValidateArgs(runAll bool, onlyRun, args []string) (string, error) {
+	if runAll == true && len(onlyRun) != 0 {
```
sudermanjr, line 9:
```suggestion
	if runAll && len(onlyRun) != 0 {
```

sudermanjr (changes requested):

PR #303, opened 2021-02-15
Set default namespace and repository on opening course file.

### LINKED ISSUE #301: Populate Default namespace and repository on initial unmarshal of course file
When we read in the course file, it will be much easier to just fill our the default namespace and repository on every release that doesn't have one, rather than handling that later.

sudermanjr:
Fixes #301

### DIFF
--- a/pkg/course/course.go
+++ b/pkg/course/course.go
@@ -244,6 +244,9 @@ func OpenCourseV2(fileName string) (*FileV2, error) {
 		return nil, err
 	}
 
+	courseFile.populateDefaultNamespace()
+	courseFile.populateDefaultRepository()
+
 	if courseFile.SchemaVersion != "v2" {
 		return nil, fmt.Errorf("unsupported schema version: %s", courseFile.SchemaVersion)
 	}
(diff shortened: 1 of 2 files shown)


pkg/course/course.go
```diff
@@ -264,3 +267,35 @@ func OpenCourseV1(fileName string) (*FileV1, error) {
 
 	return courseFile, nil
 }
+
+// This will populate the default namespaces in each release
+// if the release does not have a namespace
```
sudermanjr, line 272:
```suggestion
// populateDefaultNamespace sets the default namespace in each release
// if the release does not have a namespace. If the DefaultNamespace is blank, simply returns
```

pkg/course/course.go
```diff
@@ -264,3 +267,35 @@ func OpenCourseV1(fileName string) (*FileV1, error) {
...
+		if release.Namespace == "" {
+			klog.V(5).Infof("setting the default namespace of %s on release %s", f.DefaultNamespace, releaseName)
+			release.Namespace = f.DefaultNamespace
+			f.Releases[releaseName] = release
+		}
+	}
+}
+
+// This will populate the default repository in each release
+// if the release does not have a repository specified
```
sudermanjr, line 288:
```suggestion
// populateDefaultRepository sets the default repository in each release
// if the release does not have a repository specified. If the DefaultRepository is blank, simply returns.
```

PR #327, opened 2021-03-03
Fix Dependency update path

### LINKED ISSUE #325: Dependancy update fails when not in the same directory as the course file
When a chart has dependancies and you execute `reckoner update path/to/course.yml -o chart-with-dependancies` an error is thrown similar to
```
2021-03-02 15:37:53 C02DG0SRMD6R root[64339] ERROR ERROR: update Failed on postgres-proxy
2021-03-02 15:37:53 C02DG0SRMD6R root[64339] ERROR Stopping 'update' for chart due to an error! Some of your requested actions may not have been completed!
2021-03-02 15:37:53 C02DG0SRMD6R root[64339] ERROR Error: found in Chart.yaml, but missing in charts/ directory: ingress-nginx
2021-03-02 15:37:53 C02DG0SRMD6R root[64339] ERROR Not running Course post_install hook due to a chart install error!
```

When the same command is run but from the same directory `reckoner update course.yml -o chart-with-dependancies`
It download the dependancies just fine.

ejether:
This PR fixes an issue with dependency update for a local chart when the `reckoner` command is not run in the same directory as the `course.yml` file. The update was being excluded because thecode checking if a chart existed was relative to the command directory and not the the `course.yml`. This changes the chart path check to be absolute _and_ uses the `course.yml` location as it's starting point.

Fixes #325

### DIFF
--- a/reckoner/chart.py
+++ b/reckoner/chart.py
@@ -203,13 +203,23 @@ class Chart(object):
         """ Update the course chart dependencies """
         if self.config.dryrun:
             return True
-        logging.debug("Updating chart dependencies: {}".format(self.repository.chart_path))
-        if os.path.exists(self.repository.chart_path):
-            try:
-                response = self.helm.dependency_update(self.repository.chart_path)
-                logging.debug(response.stderr + "\n" + response.stdout)
-            except ReckonerCommandException as error:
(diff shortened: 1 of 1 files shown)


sudermanjr (approved):
awesome. thanks

PR #330, opened 2021-03-08
Adding init hook

ejether:
Adding a hook that runs immediately after parsing the hooks from the course file. This means it will be run before _any_ of the commands instead of just the `plot` command. I think this is useful so that one can use a hook to initialize kubectl or authenticateinto a cluster for later commands.


docs/README.md
```diff
@@ -79,6 +82,8 @@ The `charts` block in your course define all the charts you'd like to install an
     The version of the chart in the remote chart repository (NOTE: When using a git repository, this is translated into the `ref` in the git repository, either commit SHA or tag name)
 - `hooks` _(object)_
     Allows `pre_install` and `post_install` which can be a string or a list of strings to execute in the shell around your chart installation (execution working directory is relative to your course file)
+     - `pre_install` runs before the chart is installed/upgraded. Only active for the `plot` subcommand.
```
sudermanjr, line 85: Is this indentation off?

ejether: I don't think so. It looks fine in the file... but if you want a format changed I will do so.

sudermanjr (approved):

PR #335, opened 2021-03-10
 Fixing Strong Typing Test

ejether:
**NOTICE** 
this just fixes the tests because the expected type does not match the detected type.
I flipped the var name so the expected matches the detected.

I don't know if this is because typing has changed or if the test never worked as intended.
_Please review with care._


sudermanjr (approved):
Seems fine to me. Not sure why it would have changed. It is possible I missed it in my parallelization PR and it never worked though

sudermanjr:
Doesn't seem to have worked. :-(

sudermanjr (approved):

PR #498, opened 2021-12-02
[RD-34] Fix default namespace management

lucasreed:
Because we already populate the default namespace for a given release during the course file parse there's no need to try and create it here. It will be created in the forloop below the removed code.

This should fix some of the e2e tests because it was attempting to create the default namespace twice before.


sudermanjr:
I'm not sure I understand why this is necessary. Would you mind adding a description?

sudermanjr (approved):

PR #500, opened 2021-12-07
get-manifests command and diff command RD-33 RD-31

lucasreed:
- use colored text output instead of klog for default log level. Still use klog for any error formatting or debug logging.
- update dependencies
- ensure package docstring with license is added to all go files
- add get-manifests command and functionality
- add diff command functionality
- add some tests
- add the ability to turn colorized output off


cmd/root.go
```diff
@@ -133,9 +133,10 @@ var getManifestsCmd = &cobra.Command{
 		}
 		manifests, err := client.GetManifests()
 		if err != nil {
-			klog.Fatal(err)
+			color.Red(err.Error())
+			os.Exit(1)
```
lucasreed, line 137: Doing this to match the behavior of python reckoner. Our tests expect exit code 1 and `klog.Fatal()` exits with code 255

sudermanjr (changes requested):
Some functional testing revealed a couple usability issues that we should address. I don't think they have to be in this PR necessarily, so LMK if you want to push them to another PR

Functionally, this is fantastic and makes a lot of huge improvements to the golang version.

First, if there is no diff, we get no output at all:

```
13s ▶ ~/go/src/github.com/fairwindsops/reckoner/reckoner-go diff course.yml   -o aws-iam-authenticator
WARNING: this course file was automatically converted from v1 to v2 at runtime - to convert the file permanently, run "reckoner convert -i course.yml"
```

Second, when doing multiple releases there's no divider or indicator which on is which:

```
15s ▶ ~/go/src/github.com/fairwindsops/reckoner/reckoner-go diff course.yml  -o metrics-server  -o aws-iam-authenticator
WARNING: this course file was automatically converted from v1 to v2 at runtime - to convert the file permanently, run "reckoner convert -i course.yml"

Deployment "metrics-server"

                  command:
                    - metrics-server
                  image: docker.io/bitnami/metrics-server:0.4.2-debian-10-r0
-                 imagePullPolicy: IfNotPresent
+                 imagePullPolicy: Always
                  livenessProbe:
                    failureThreshold: 3
                    httpGet:
                        path: /livez
```

sudermanjr (approved):
Awesome.

PR #501, opened 2021-12-08
[RD-35] Add update command

lucasreed:
- move some diff functionality to a new function to only get individual diff string
- no need to return a string from the client.Plot() method because we never returned anything from it
- add update command to only run upgrades on releases that have diffs

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -66,6 +66,7 @@ func init() {
 	rootCmd.AddCommand(diffCmd)
 	rootCmd.AddCommand(lintCmd)
 	rootCmd.AddCommand(getManifestsCmd)
+	rootCmd.AddCommand(updateCmd)
 
 	klog.InitFlags(nil)
 	pflag.CommandLine.AddGoFlag(flag.CommandLine.Lookup("v"))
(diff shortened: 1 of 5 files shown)


cmd/root.go
```diff
@@ -216,6 +216,23 @@ var convertCmd = &cobra.Command{
...
+	Long:    "Only install/upgrade a release if there are changes.",
+	PreRunE: validateArgs,
+	Run: func(cmd *cobra.Command, args []string) {
+		client, err := reckoner.NewClient(courseFile, version, runAll, onlyRun, true, dryRun, createNamespaces, courseSchema)
+		if err != nil {
+			klog.Fatal(err)
+		}
+		err = client.Update()
+		if err != nil {
+			klog.Fatal(err)
```
sudermanjr, line 231: Did we want to stick with the os.Exit pattern and red error logs rather than fatal?

sudermanjr (approved):
LGTM. non-blocking question inline

cmd/root.go
lucasreed: Yes, we will. I was going to do a logging PR once some main functionality was done. I probably should have just started at that state for this command. I'll fix it.

PR #505, opened 2022-01-03
[go] parse environment variables

lucasreed:
Also changed in this PR:

* make conversion from v1 to v2 schema a private function since everything should happen with the OpenV2 function.
* changed all klog.Fatal() calls in the cmd package to red error output with exit code of 1 instead
* `reckoner convert` command no longer calls the ConvertV1toV2 function and instead simply uses OpenV2 function.

### DIFF
--- a/pkg/course/course.go
+++ b/pkg/course/course.go
@@ -18,6 +18,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"strings"
 
 	"github.com/fatih/color"
 	"github.com/xeipuuv/gojsonschema"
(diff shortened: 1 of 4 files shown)


pkg/course/course.go
```diff
@@ -261,14 +263,25 @@ func OpenCourseV2(fileName string, schema []byte) (*FileV2, error) {
...
+	data, _ = yaml.Marshal(courseFile)
+	data, err = parseEnv(data)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse env variables: %v", err)
+	}
+	err = yaml.Unmarshal(data, courseFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unmarshal file after parsing env vars: %s", err.Error())
+	}
+
```
lucasreed, line 284: The reason we are doing this unmarshal -> marshal -> unmarshal again is because if we parsed environment variables before the first unmarshal, it would also error out on any YAML comments that mention an environment variable. When we do the first unmarshal, all comments are stripped so we can marshal back to bytes, populate all the env vars and then unmarshal into our struct again.

sudermanjr: That's actually really cool. That was fairly annoying in the python version at one point.

sudermanjr (approved):

PR #507, opened 2022-02-09
RD-30 go git repository management

### LINKED ISSUE #299: [Go version] Implement git repository management
Reckoner can pull down charts from a git repository and use them as local charts. Defined in the repositories block as `git: <URL>`

lucasreed:
This PR fixes #299

## Checklist
* [x] I have signed the CLA
* [ ] I have updated/added any relevant documentation

Allow a user to specify a git repository instead of a traditional helm
repo

This basically works the same way as the python version,
cloning/checking out the git repo in the local helm cache, grabbed from
`helm env`. The main difference is that the go-git library does not
support sparse checkouts so it may take longer to pull down large
repositories with multiple charts.

I don't love how big the `setupGitRepoPath` function got, so I'm
definitely looking for feedback on that.


pkg/course/course.go
```diff
@@ -92,6 +92,7 @@ type NamespaceConfig struct {
 
 // Release represents a helm release and all of its configuration
 type Release struct {
+	gitClonePath *string
```
sudermanjr, line 95: Let's export this field and then we can drop the `GitClonePath()` method.

pkg/course/course.go
```diff
@@ -397,6 +398,23 @@ func (f FileV2) validateJsonSchema(schemaData []byte) error {
 	return nil
 }
 
+// SetGitClonePath allows the caller to set the path of a locally stored chart from a git repository already cloned
+func (r *Release) SetGitClonePath(path string) error {
```
sudermanjr, line 402: Let's call this when we read in the course file, in the NewClient() function from the reckoner package.

pkg/reckoner/plot.go
```diff
@@ -124,11 +135,59 @@ func (c Client) TemplateRelease(releaseName string) (string, error) {
...
+		return fmt.Errorf("Error fetching git repository %s - %s", clonePath, err)
+	}
+
+	hash, err := repo.ResolveRevision(plumbing.Revision(release.Version))
+	if errors.Is(err, plumbing.ErrReferenceNotFound) {
+		hash, err = repo.ResolveRevision(plumbing.Revision(fmt.Sprintf("origin/%s", release.Version)))
+	}
+	if err != nil {
+		return fmt.Errorf("Error resolving git revision %s - %s", release.Version, err)
+	}
```
sudermanjr, line 174: This could maybe be a separate function

sudermanjr (changes requested):

sudermanjr (approved):

PR #530, opened 2022-02-28
Refactoring go version and many fixes for e2e tests

lucasreed:
## Checklist
* [x] I have signed the CLA

To fix most e2e tests for the go rewrite by adding the git repository
management feature and fixing some bugs that were uncovered by these
tests.

Also removed all python testing from the circleci config and temporarily
commented out the release workflow until we are ready to release the go
version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,35 +17,8 @@ version: 2.1
 orbs:
   rok8s: fairwinds/rok8s-scripts@11
 
-executors:
-  python-3-7:
-    docker:
-      - image: circleci/python:3.7
-  python-3-8:
-    docker:
-      - image: circleci/python:3.8
-  python-3-9:
-    docker:
-      - image: cimg/python:3.9
-
 references:
-  e2e_configuration_python: &e2e_configuration_python
-    pre_script: end_to_end_testing/pre_python.sh
-    script: end_to_end_testing/run_python.sh
-    command_runner_image: quay.io/reactiveops/ci-images:v11-buster
-    enable_docker_layer_caching: false
-    store-test-results: /tmp/test-results
-    kind_version: 0.10.0
-    requires:
-      - build-3-7
-      - build-3-8
-      - build-3-9
-    filters:
-      branches:
-        only: /.*/
-      tags:
-        ignore: /.*/
-  e2e_configuration_go: &e2e_configuration_go
+  e2e_configuration: &e2e_configuration
     attach-workspace: true
     workspace-location: /
     pre_script: end_to_end_testing/pre_go.sh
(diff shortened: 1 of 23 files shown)


sudermanjr (changes requested):

pkg/course/course.go
```diff
@@ -54,21 +63,24 @@ type FileV2 struct {
 	// NamespaceMgmt contains the default namespace config for all namespaces managed by this course.
 	NamespaceMgmt struct {
 		// Default is the default namespace config for this course
-		Default NamespaceConfig `yaml:"default" json:"default"`
+		Default *NamespaceConfig `yaml:"default" json:"default"`
 	} `yaml:"namespace_management,omitempty" json:"namespace_management,omitempty"`
 	// Releases is the list of releases that should be maintained by this course file.
-	Releases []*Release `yaml:"releases" json:"releases"`
+	Releases []*Release `yaml:"releases,omitempty" json:"releases,omitempty"`
```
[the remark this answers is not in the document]
lucasreed: This is on purpose because it will populate this in the struct and then jsonschema validation passes even if releases doesn't exist.

pkg/course/course.go
```diff
@@ -258,10 +271,10 @@ func OpenCourseV2(fileName string, schema []byte) (*FileV2, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	err = yaml.Unmarshal(data, courseFile)
 	if err != nil {
-		return nil, fmt.Errorf("failed to unmarshal file: %s", err.Error())
+		klog.V(3).Infof("failed to unmarshal file: %s", err.Error())
```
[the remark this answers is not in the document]
lucasreed: Because I standardized on a single error message via the `SchemaValidationError` error type I provide more context to the underlying issue with a higher verbosity logging.

.circleci/config.yml
```diff
@@ -186,89 +76,75 @@ jobs:
...
-            - test-go
+            - test
           filters:
             tags:
               ignore: /.*/
             branches:
               only: /.*/
       - rok8s/kubernetes_e2e_tests:
-          name: "End-To-End Kubernetes 1.19 - Python"
+          name: "End-To-End Kubernetes 1.19.7"
```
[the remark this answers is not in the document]
lucasreed: Yep, I'll do that.

.circleci/config.yml
```diff
@@ -17,114 +17,26 @@ version: 2.1
...
+  e2e_configuration: &e2e_configuration
     attach-workspace: true
     workspace-location: /
     pre_script: end_to_end_testing/pre_go.sh
     script: end_to_end_testing/run_go.sh
     command_runner_image: quay.io/reactiveops/ci-images:v11-buster
     enable_docker_layer_caching: false
     store-test-results: /tmp/test-results
-    kind_version: 0.10.0
+    kind_version: 0.11.1
```
sudermanjr, line 29: This should be the default now I think
```suggestion
```

sudermanjr (changes requested):

pkg/course/course.go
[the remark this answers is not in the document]
sudermanjr: 👍🏼

pkg/course/course.go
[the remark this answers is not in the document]
sudermanjr: Ahhh, good call.

.circleci/config.yml
lucasreed: Oh nice!

cmd/root.go
```diff
@@ -165,6 +174,9 @@ var diffCmd = &cobra.Command{
 			color.Red(err.Error())
 			os.Exit(1)
 		}
+		if client.Errors > 0 {
+			os.Exit(1)
+		}
```
lucasreed, line 179: Add this to the other places where we can continue-on-err

pkg/course/course.go
```diff
@@ -313,24 +321,78 @@ func OpenCourseV1(fileName string) (*FileV1, error) {
...
+	if err := value.Decode((*FileV1Unmarshal)(f)); err != nil {
+		return err
 	}
-	*cl = make([]ReleaseV1, len(value.Content)/2)
-	for i := 0; i < len(value.Content); i += 2 {
-		var res = &(*cl)[i/2]
-		if err := value.Content[i].Decode(&res.Name); err != nil {
-			return err
+	return nil
+}
```
lucasreed, line 333: Add comment and also add FileV2 UnmarshalYAML

pkg/course/course.go
```diff
@@ -374,7 +436,22 @@ func (f *FileV2) populateEmptyChartNames() {
...
+		f.NamespaceMgmt.Default = &emptyNamespaceMgmt
+	}
+	for releaseIndex, release := range f.Releases {
+		if release.NamespaceMgmt == nil {
+			klog.V(5).Infof("using default namespace management for release: %s", release.Name)
+			release.NamespaceMgmt = f.NamespaceMgmt.Default
+			f.Releases[releaseIndex] = release
+		}
+	}
+}
```
lucasreed, line 452: Check about merging vs. overwriting namespace_management and which takes precedence in the docs.

pkg/course/course.go
```diff
@@ -409,19 +486,18 @@ func (r *Release) SetGitPaths(clonePath, subPath string) error {
 	return nil
 }
 
-func parseEnv(data []byte) ([]byte, error) {
+func parseEnv(data string) (string, error) {
 	dataWithEnv := os.Expand(string(data), envMapper)
```
lucasreed, line 490:
```suggestion
	dataWithEnv := os.Expand(data, envMapper)
```

pkg/reckoner/plot.go
```diff
@@ -20,16 +20,15 @@ import (
 	"os"
 	"strings"
 
-	"gopkg.in/yaml.v3"
-	"k8s.io/klog/v2"
+	"gopkg.in/yaml.v2"
```
lucasreed, line 23: can we change this back?

pkg/reckoner/plot.go
```diff
@@ -20,16 +20,15 @@ import (
...
 
 	"github.com/fairwindsops/reckoner/pkg/course"
 	"github.com/fatih/color"
 	"github.com/thoas/go-funk"
 )
 
 // Plot actually plots the releases
-func (c Client) Plot() error {
+func (c *Client) Plot() error {
```
lucasreed, line 31: Confirm why this was changed.

pkg/reckoner/update.go
```diff
@@ -21,17 +21,21 @@ import (
 
 func (c *Client) Update() error {
 	updatedReleases := []*course.Release{}
-	for i, r := range c.CourseFile.Releases {
-		thisReleaseDiff, err := c.diffRelease(r.Name, r.Namespace)
+	for i, release := range c.CourseFile.Releases {
```
lucasreed, line 24:
```suggestion
	for idx, release := range c.CourseFile.Releases {
```

For readability and then change the instances of `i` below.

lucasreed (commented):
Things found when reviewing with Andy

cmd/root.go
lucasreed: I already did this actually. I was confused because the `reckoner.NewClient` function requires it as an arg even on commands that don't make use of that functionality.

pkg/reckoner/plot.go
lucasreed: I changed this because `c.Continue()` uses a pointer to a client object. It doesn't seem to be a syntax error if we change it back to a non-pointer but I'm not really sure the repercussions.

sudermanjr: Technically, it only needs to be a pointer if you're going to modify it inside the function. I've seen a general convention to just make a it a pointer though, so it's probably fine?

sudermanjr (approved):

PR #536, opened 2022-03-03
add support for specifying course file via environment variable

### LINKED ISSUE #503: allow reckoner to be configured via environment variables
Cobra is currently used to take in CLI arguments. I think it would be very helpful if `reckoner` picked up an environment variable on my machine that I pre-populate -- say for example `${reckoner_course_file}`.
Currently I just type the variable name each time I want to run `reckoner`. That could be avoided with this proposed enhancement.

I'm unsure if adding more arguments than just the path to the course file makes sense or not. Please evaluate the current Cobra commands and decide if supporting more commands via environment variables seems reasonable.

intrand:
This PR fixes #503

To be able to configure the path to course.yml via environment variable. This avoids passing a path on the command line at every invocation, which will help with automation and possibly usability in single-cluster environments.

In addition, I've added a default path of `course.yml` to the courseFile. This should further help reduce typing if you're in the same directory as a course.yml (assuming it's named that).

cmd/root.go now handles course file setting precedence. This should be global to all instances of setting courseFile, not just any particular subcommand (eg, it applies to `plot`, `convert`, `template`, etc).

Using [viper](https://github.com/spf13/viper) to handle configuration overrides at [various levels](https://github.com/spf13/viper#why-viper). However, that's a much larger change.


pkg/reckoner/validation_test.go
```diff
@@ -39,76 +39,124 @@ func TestValidateArgs(t *testing.T) {
...
+			want:    "testdata/course.yaml",
+			wantErr: false,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got, err := ValidateArgs(tt.args.runAll, tt.args.onlyRun, tt.args.args)
+			// note: I do not like repeating logic from cmd/root.go in
+			// this test. instead, or additionally, perhaps we should
+			// test root.go directly to gain its effects
```
lucasreed, line 133: I don't know if you can get around that unless you move that logic all into this package and test it fully here. Even if you test root.go you'll need to get a course.yml to work with in this test.

Also, lets remove this comment block.

lucasreed: Thinking about this, it seems to be it would be easier to just move these validating functions into the cmd package where they really matter. I'll add that to my top level review comment.

lucasreed (changes requested):
It strikes me as I read this that validating the args probably never belonged in the `reckoner` package. Since we are, as the name implies, validating command line arguments, I think that all belongs in the `cmd` package. That would probably make testing easier and decrease the size of the very large reckoner package since this code doesn't really make sense there.

Once moved into the `cmd` package they don't need to be capitalized (exported) since they can be private functions at that point.

I have a couple other comments in-line that are less critical.

pkg/reckoner/validation.go
```diff
@@ -20,22 +20,26 @@ import (
 )
 
 // ValidateArgs validates that the correct arguments were passed and returns the name of the course file
-func ValidateArgs(runAll bool, onlyRun, args []string) (string, error) {
+func ValidateArgs(runAll bool, onlyRun []string, args []string) error {
```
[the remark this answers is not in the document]
intrand: Reverted this change as suggested.

pkg/reckoner/validation_test.go
intrand: Moved code & removed comment.

cmd/validation.go
```diff
@@ -0,0 +1,73 @@
...
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+	"k8s.io/klog"
```
lucasreed, line 23:
```suggestion
	"k8s.io/klog/v2"
```

We recently switched to using klog v2 everywhere

lucasreed (changes requested):
One small fix in klog version so we aren't re-introducing the old version. Other than that it all looks great, thanks!

cmd/validation.go
lucasreed: Once you change this, make sure you run `go mod tidy` as well.

intrand:
All set!

PR #537, opened 2022-03-04
[go] fix kubecontext checking

lucasreed:
## Checklist
* [x] I have signed the CLA

Make use of the toplevel chart key `context` so we use the configured kubecontext. If it does not exist in the kubeconfig, we should bail out.


sudermanjr (approved):
Well that was easy

PR #559, opened 2022-03-31
go.mod: update kevinburke/ssh_config to latest version

kevinburke:
Previously if you tried to read a HostName in a config file that
looked like this:

```
Host github
    HostName github.com        # This is the host for code review
```

DefaultUserSettings.Get("HostName") would return "github.com        ",
which I think is unintuitive and unexpected.

This behavior is fixed in v1.2 which would return "github.com" in the
above example.


sudermanjr:
I rolled this into another PR. Thanks!!

PR #566, opened 2022-04-08
Fix to the env var escaping

sudermanjr:
## Checklist
* [x] I have signed the CLA

Previous logic put the `$$` back, rather than replacing it with a single `$`

One character removed in the output of the check, and fixed the test.


lucasreed (approved):
Love it when it's a simple fix! 🎉

PR #573, opened 2022-04-20
remove python version completely

intrand:

No longer maintain nor offer the legacy python version of reckoner. Any references can be made by looking at a previous release/commit.

Removed all of the python stuff :)

We've already decided to drop it


lucasreed (commented):
This is a great start, I think there are some python specific things in the [docs/ directory](https://github.com/FairwindsOps/reckoner/tree/master/docs) too if you wouldn't mind looking through those (mostly around installation I think).

intrand:
> This is a great start, I think there are some python specific things in the [docs/ directory](https://github.com/FairwindsOps/reckoner/tree/master/docs) too if you wouldn't mind looking through those (mostly around installation I think).

Sure thing.

lucasreed (approved):
Woot! Thanks 🎉

PR #578, opened 2022-04-22
Don't parse env on convert command

### LINKED ISSUE #577: reckoner convert should not parse environment variables
### What happened?

I converted a course file that has env vars in it, and it replaced them in the converted version

### What did you expect to happen?

Leave the env-vars the way they are in the conversion command. This will prevent leaking secrets on accident

### How can we reproduce this?

create a v1 course file with an env var in it. Run `reckoner convert course.yaml`

### Version

v6.0.0-rc.4

### Search

- [X] I did search for other open and closed issues before opening this.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

### Additional context

_No response_

sudermanjr:
This PR fixes #577

## Checklist
* [x] I have signed the CLA

When running `reckoner convert`, we shouldn't parse environment variables

Added a package variable to disable env parsing

There's likely a cleaner way to do this, I'm not overly happy about using a global for this, but due to the extreme ways in which we have to have to handle yaml, this may be necessary

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -276,6 +276,7 @@ var convertCmd = &cobra.Command{
 		return validateCobraArgs(cmd, args)
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		course.ParseEnv = false
 		newCourse, err := course.OpenCourseFile(courseFile, courseSchema)
 		if err != nil {
 			color.Red(err.Error())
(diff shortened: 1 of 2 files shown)


lucasreed (approved):
I can't think of a better way to do this either without introducing a new struct that is only for env parsing and even then I'm not 100% sure that would be pretty either.

It's really striking how much of this course package revolves around unmarshaling and environment variable parsing. I'm not sure there's a better way, but it still feels weird.

PR #584, opened 2022-04-26
remove pip from dependabot

intrand:
I must've missed this during the python removal PR. Whoops.


sudermanjr:
I think this is managed by the centralized tf. Likely should add a comment.

intrand:
Will update with terraform

PR #585, opened 2022-04-26
make shell test work universally (hopefully)

intrand:

make the tests for shells work universally. they were failing on some Ubuntu-based distros at the least.

The original test just made sure the result of `newShellExecutor()` matched an object hardcoded in `Test_newShellExecutor()`.
Now we independently test the attributes of the resulting object against what we want from the test, using a regex to test `got.Executable` and EqualValues for `got.Args`, respectively.

We talked about supporting a list of possible places to expect `echo` on a distribution, but what if the system configuration is valid and functional, but we didn't include support? Ultimately we decided to go with a regex so we likely won't have to look at this again for a long, long time.

### DIFF
--- a/pkg/course/shellSecrets_test.go
+++ b/pkg/course/shellSecrets_test.go
@@ -15,6 +15,7 @@
 package course
 
 import (
+	"regexp"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


sudermanjr (approved):
👍

PR #596, opened 2022-05-31
support argocd application generation via course file

intrand:
This PR fixes #

## Checklist
* [x] I have signed the CLA

Make reckoner more friendly to gitops users, including ourselves. This implements only Applications for ArgoCD currently. More support may come in the future.

Custom structs to support a subset of ArgoCD Application custom resources. We could've imported the entire struct tree from upstream, but it relies upon kubernetes kinds, which we did not want to introduce as a dependency to reckoner.


pkg/course/argocd.go
```diff
@@ -0,0 +1,53 @@
+package course
+
+// import (
+// 	argoAppv1alpha1 "github.com/argoproj/argo-cd/v2/pkg/apis/application/v1alpha1"
+// 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+// )
```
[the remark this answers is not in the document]
sudermanjr: Not sure if this was just from testing the import, or intentionally left to show how to convert later if necessary. If the latter, then a comment explaining would be better than removing.

pkg/reckoner/plot.go
```diff
@@ -138,6 +138,19 @@ func (c Client) TemplateAll(templateOutputDir string) (fullOutput string, err er
...
+	// and a warning when not found in the config file. This
+	// maybe could use some improvement
+	if len(c.CourseFile.GitOps.ArgoCD.Spec.Source.RepoURL) > 0 {
+		err = c.WriteArgoApplications(templateOutputDir)
+		if err != nil {
+			color.Red(err.Error())
+			os.Exit(1)
+		}
+	}
+
```
sudermanjr, line 153: I'm noticing that this only runs when we `template all` - but if I'm not mistaken, we can still do --output-dir when we only template a single chart. 

I think this should be triggered closer to the actual template function, perhaps in the TemplateRelease function, rather than here. What are your thoughts?

pkg/reckoner/argocd.go
```diff
@@ -0,0 +1,112 @@
...
+
+	for _, release := range c.CourseFile.Releases {
+		// generate an argocd application resource
+		app, err := generateArgoApplication(*release, c.CourseFile)
+		if err != nil {
+			return err
+		}
+
+		// generate name of app file
+		appOutputFile := appsOutputDir + "/" + app.Metadata.Name + ".yaml"
```
sudermanjr, line 93: Can we please add toLower here?

sudermanjr (changes requested):

pkg/course/argocd.go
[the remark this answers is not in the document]
intrand: added a comment block to explain why the commented imports are present, and what to do about them in the future when a final decision is made

pkg/reckoner/plot.go
intrand: Good catch! I can't believe I didn't notice this during development/testing. Thanks very much for pointing that out. I've moved it to the appropriate function (`TemplateRelease()`) where it should've been all along :)

pkg/reckoner/argocd.go
intrand: done

intrand:
while I was demoing this to @sudermanjr, we noticed several things that we feel could be improved upon. A couple of them made it into the formal review, while the rest I kept notes on. There are more commits coming. I'll re-request review from everyone when I believe I've satisfied the notes from our meeting.

sudermanjr (changes requested):
Seems like these changes broke the tests. Specifically test 10 with git charts. Not sure what's up there.

sudermanjr:
Opened https://github.com/FairwindsOps/reckoner/pull/597 to fix the tests

pkg/reckoner/plot.go
```diff
@@ -122,27 +122,20 @@ func (c Client) TemplateAll(templateOutputDir string) (fullOutput string, err er
...
-
 		fullOutput = fullOutput + out
 	}
+
 	return fullOutput, nil
 }
 
 // TemplateRelease does the same thing as TemplateAll but only for one release
-func (c Client) TemplateRelease(releaseName string) (string, error) {
+func (c Client) TemplateRelease(releaseName string, templateOutputDir string) (string, error) {
```
sudermanjr, line 138: I think for a future PR, it would be good to make templateOutputDir a field of the client struct, rather than passing it into these functions separately, since it is a global configuration passed in via the CLI. 

Non-blocking for this PR

pkg/reckoner/plot.go
```diff
@@ -157,6 +150,27 @@ func (c Client) TemplateRelease(releaseName string) (string, error) {
...
+			color.Red(err.Error())
+			os.Exit(1)
+		}
+	}
+
+	// if the user has specified a RepoURL, they probably also
+	// specified other things. Most required things receive a default
+	// and a warning when not found in the config file. This
+	// maybe could use some improvement
+	if c.CourseFile.GitOps.ArgoCD.Spec.Source.RepoURL != "" && templateOutputDir != "" {
```
sudermanjr, line 166: Would it be better to do 

```
if _, ok := c.CourseFile.GitOps.ArgoCD ; ok {

  // Write out the argocd Applications
```

This could go hand-in hand with adding the outputDir to the client struct, and then check if it is set inside of the writeArgoApplications function.

sudermanjr (approved):
I left some non-blocking suggestions that could be done in a separate PR to clean a couple things up.

pkg/reckoner/plot.go
intrand: I'm not sure that works 🤔 please elaborate.

another idea I'll throw into the discussion: we could compare `c.CourseFile.GitOps.ArgoCD` to `course.ArgoApplication{}` and see if they (don't) match. if they don't match, something was modified via the course file. if they match, nothing has been given.

🤔 we could probably come up with some other ideas, too.

sudermanjr: Your idea seems good to me, and achieves what I was trying to suggest, or as we talked about today probably good to just leave as-is.

PR #620, opened 2022-08-22
many upgrades

intrand:
## Checklist
* [x] I have signed the CLA

Handle golang & dependency upgrades

1. bump golang version to 1.18

2. `go get -u -t ./...` but reverted the `github.com/sergi/go-diff` bump because v1.2.x is causing problems

n/a


sudermanjr (approved):

sudermanjr:
Oh, you'll probably need to update goreleaser to the latest in order to support newer versions of Go

intrand:
was wondering how that'd turn out :)

